Issue #5, opened 2016-04-19
FR: Bash Autocompletion


jhedwardyang:
Tab should recommend and autocomplete for commands and options

PR #17, opened 2016-04-25
[DO NOT MERGE] Bash completion proof of concept

jhedwardyang:
This is a first iteration of #5 for proof of concept

Implementation details:
- Run `git-meta --completion` to generate the bash script. Source this (this can be automated simply by saving it to `/usr/local/etc/bash_completion.d/` later)
- In `node_modules/omelette/src/omelette.js`, comment out line 58 (`return process.exit();`). Reason: since bash completion must occur before other codepaths are run (or to be exact, should be run instead of other codepaths), omelette was written in a way that emits the events, then force calls process.exit. However, since our code uses promises, the process.exit may be called before our promise is fulfilled in the event loop. 

Some terminology I use:

- option - flag that starts with - or --
- completer - method that is called to "complete" based on a given prefix
- command - command or sub-command
- positional - positional argument

### DIFF
--- a/lib/autocomplete/autocomplete.js
+++ b/lib/autocomplete/autocomplete.js
@@ -1,0 +1,171 @@
+#!/usr/bin/env node
+/*
+ * Copyright (c) 2016, Two Sigma Open Source
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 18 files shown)


jhedwardyang:
Abandoning for now, will come back and look at this.

PR #34, opened 2016-05-23
submodule support for AST/shorthand

bpeabody:
- Also reworked writing to build tree for each commit from scratch instead of
  working from parent commit.  This change had a couple of benefits:
  * same logic could be used with root and child commits
  * no cleanup (e.g. with HEAD) issues for working dir during generation
  * don't need to actually connect to or fecth from submodules to add them
  * can create submodule references to repos/commits that don't exist
- split out the submodule config logic into a lower-level component usable by
  the test system
- logic to "render" a commit chain into a logical workdir tree

### DIFF
--- a/lib/util/repoast.js
+++ b/lib/util/repoast.js
@@ -39,6 +39,45 @@ const assert   = require("chai").assert;
 const deepCopy = require("deepcopy");
 
 /**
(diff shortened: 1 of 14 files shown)


bpeabody:
FYI: @jhedwardyang

Issue #29, opened 2016-05-10
Make `git meta close` more foolproof


abliss:
1) before closing, check for commits which have not been pushed anywhere
2) Instead of running rm -rf .git/modules/foo/bar, maybe make a backup somewhere?

bpeabody:
I don't think we should do (2). This behavior would be un-gitlike. I would say it should:

1. Warn and refuse to run if there are unpunished (or index or working directory) changes.
2. Do `rm -rf` and do not warn if the users passes the `-f` flag.

abliss:
I appreciate the Git mentality of giving the user powerful and dangerous
tools. But git also makes it a priority never to destroy committed data.
Even if you amend or rebase, the reflog is there to save you (at least
until the next gc). Can you name any git operation (even with -f) that
would immediately and permanently destroy a commit?

There doesn't seem to be any reliable way to know that all commits in a
repo exist on the remote. (Even if there were, relying on it isn't very
gitlike either, since the remote itself could become unreachable.)

If you're worried about runaway backups, maybe there's a way to plug into
git gc to do cleanup?

bpeabody:
What I mean by "un-gitlike" is:
1. Many other Git commands can remove user content, they all require some form of `-f` before they delete data.
2. Git users generally are less concerned about such loss because they commit and push often; the idea to back up uncommitted/pushed content because there might be a lot of it makes seems more like an `svn` (or even old-school `hg`) style concept.
3. Specifically, the related `git` command, `git submodule deinit` does `rm -rf` when forced.

I think the idea of making backups is also contrary to the mandate of `git-meta`, which is to do only things that could be done by basic `git` commands; I think we'd have a hard time getting this functionality accepted into Git.

I think giving them a big warning -- clearly show them what is going one -- and not making any destructive changes unless they type `--force` is more than sufficient; the user can easily back up the files on his or her own after seeing the warning.

I'm OK with the possibility of the remote being unreachable or the user doing something clearly wrong to make `remote/origin` branches appear as if the remote has been updated when it actually hasn't.

We will actually be doing a better job than `git submodule deinit`, by the way, which does not check for unpushed commits.

bpeabody:
I think we're good on this now. `close` will not remove workdir contents without `-f`, and and committed/staged changes will be preserved in `.git/modules/<submodule path>` no matter what the user does.

PR #36, opened 2016-05-24
git meta close checks for unpushed or uncommited changes

jhedwardyang:
Work for #29 

`git meta close` now requires `--force` to succeed if:
- There are unpushed changes (to any remote)
- The repository is not clean (uncommit or unstaged changes)
- There are new files

### DIFF
--- a/lib/util/close.js
+++ b/lib/util/close.js
@@ -39,6 +39,11 @@ const rimraf  = require("rimraf");
 const path    = require("path");
 const fs      = require("fs-promise");
 
+const GitUtil   = require("./git_util");
+const NodeGit   = require("nodegit");
+const Status    = require("./status");
+const UserError = require("./user_error");
+
 /**
  * Remove the modules directory for the submodule having the specified 'name'
  * in the repo located at the specified 'root'.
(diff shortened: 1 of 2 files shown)


lib/util/close.js
```diff
@@ -61,22 +66,75 @@ exports.cleanModulesDirectory = co.wrap(function *(root, name) {
...
+        const headCommit = yield subRepo.getHeadCommit();
+        const remotes = yield NodeGit.Remote.list(subRepo);
+
+        const checkUnpushedCommits = co.wrap(function *(remote) {
+            const unpushed = yield GitUtil.listUnpushedCommits(
+                subRepo,
+                remote,
+                headCommit.toString());
+            if (0 === unpushed.length) {
+                allPushed = true;
```
abliss: This appears to check that there exists one remote which has all the commits, rather than checking that every commit is on at least one remote.

lib/util/close.js
```diff
@@ -61,22 +66,75 @@ exports.cleanModulesDirectory = co.wrap(function *(root, name) {
...
+        const submodule = yield NodeGit.Submodule.lookup(repo, submoduleName);
+        const subRepo = yield submodule.open();
+
+        const errorMessages = [];
+
+        // Determine if all of the commits have been pushed to some remote
+
+        let allPushed = false;
+
+        const headCommit = yield subRepo.getHeadCommit();
```
abliss: Don't we need to check this for every head in the repo?

lib/util/close.js
```diff
@@ -61,22 +66,75 @@ exports.cleanModulesDirectory = co.wrap(function *(root, name) {
...
+        const remotes = yield NodeGit.Remote.list(subRepo);
+
+        const checkUnpushedCommits = co.wrap(function *(remote) {
+            const unpushed = yield GitUtil.listUnpushedCommits(
```
abliss: listUnpushedCommits doesn't do a fetch. So if I make a commit, and run "git push origin HEAD:foo/bar", then my commit is pushed, but none of my locally-stored remote-tracking branches will know about it. So "git meta close" will throw a spurious warning?

lib/util/close.js
jhedwardyang: Yes, good catch - I was meaning to make this fix

lib/util/close.js
jhedwardyang: GitUtil.listUnpushedCommits checks every ref, the headCommit is only used if we find no remote.

lib/util/close.js
bpeabody: I think that `git push origin HEAD:foo/bar` will update `remotes/origin/foo/bar`.

I do not think that `close` (or `push`) should do a fetch first. Related Git commands to send things upstream do not work that way.

jhedwardyang: @bpeabody confirmed.
```
$ echo >> hi
$ git add .
$ git commit -m "A"
$ git log -n 2
commit 4a54f265af57d8db4701f262818a730dd95ee904
Author: Edward Yang <jh.edwardyang@gmail.com>
Date:   Tue May 24 11:44:31 2016 -0400

    A

commit 5d4ba257ef19171240fc398ff92881ab2ab634c8
Author: Edward Yang <jh.edwardyang@gmail.com>
Date:   Tue May 24 11:35:02 2016 -0400

    Initial commit

$ git push origin HEAD:foo
Counting objects: 3, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (2/2), done.
Writing objects: 100% (3/3), 269 bytes | 0 bytes/s, done.
Total 3 (delta 0), reused 0 (delta 0)
To http://github.com/jhedwardyang/test1
 * [new branch]      HEAD -> foo

$ ls .git/refs/origin/remotes/origin
HEAD	foo
$ cat .git/refs/origin/remotes/origin/foo
4a54f265af57d8db4701f262818a730dd95ee904
```

abliss: That only works if a "git fetch origin" _would_ fetch that commit. I.e., the remote ref must be listed in remote.origin.fetch.

If you set remote.origin.fetch to be +refs/heads/master:refs/remotes/origin/master rather than the default which is usually +refs/heads/*:refs/remotes/origin/* , then you'll find that there is no local record of the push.

Alternately, with your current config, try pushing to something that is not under refs/heads -- for example, git push HEAD:refs/foo/bar . There will be no local record of the push.

bpeabody: The decisions we make based on unpushed commits (which autocorrect wanted
to change to "unpunished") are best-guess; there are ways to circumvent it.
Generally speaking, we're looking for commits that haven't been pushed from
your repository; if they got there some other way, we don't expect to know
about them.

The second point you bring up: "pushing to something that is not under
refs/heads ... There will be no local record of the push." is expected
behavior. `refs/heads` is the standard location for branches to live and
other git-based tools will also be unable to detect pushes to targets
outside `heads`. I think workflows involving non `heads` targets should be
clear that commits pushed to such locations may trigger actions but don't
actually live in the repository in a first-class way.

bpeabody:
oops, we should have had CI on this one; broke some tests

PR #37, opened 2016-05-24
Backups .git/modules/submodule

jhedwardyang:
Work for #29 

Looking for some feedback:

- Currently just kills the old backup directory and replaces it.
- .git/metabackups might not be the best directory (perhaps .gitmeta/backups?)
- In the future perhaps extend this into a util/backup?

### DIFF
--- a/lib/util/close.js
+++ b/lib/util/close.js
@@ -40,18 +40,29 @@ const path    = require("path");
 const fs      = require("fs-promise");
 
 /**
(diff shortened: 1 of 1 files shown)


bpeabody:
Check out my comments on the "backups" directory; I don't think we should do backups

jhedwardyang:
Abandoning this PR since we plan on investigating deinit, init and changing open if necessary instead.

Issue #38, opened 2016-05-24
FR: `git meta status` should print every open submodule


abliss:
I think `git meta status --summary` is supposed to do this, but (a) that doesn't work, and (b) it's counterintuitive that the _summary_ is _more_ verbose. I think our users will expect to see a list of every codebase from `git meta status` and the current "brief" behavior should be an option.

bpeabody:
I think 'summary' may be the wrong word for that output. We do need an easy way to see all open sub modules, maybe as a separate command or an option to status. 

The 'status' command, itself, gives information about changes to potentially be committed; I don't think it should show unchanged open submodules any more than it should show unchanged paths. It's modeled after 'git status', which has the following as the first paragraph in its documentation:

Displays paths that have differences between the index file and the current HEAD commit, paths that have differences between the working tree and the index file, and paths in the working tree that are not tracked by Git (and are not ignored by gitignore[5]). The first are what you would commit by running git commit; the second and third are what you could commit by running git add before running git commit.

bpeabody:
My proposal:
- change `--summary` to `--list-submodules`
- add a `-a` or `--all` option to `git meta status` that will list all open submodules, even the ones without modifications

bpeabody:
Or we could do what @jhedwardyang did in this PR https://github.com/twosigma/git-meta/pull/39#event-670908128, and leave `status` alone but make the `--summary` command print only open submodules by default. I'm going to recommend changing that to `--list-submodules` while we're at it.

jhedwardyang:
Closing this since it was fixed.

PR #39, opened 2016-05-24
Fix for missing shas in status, --verbose option to summary

jhedwardyang:
For #38 

Changes:

- Fixes the --summary option when shas are missing
- Changes --summary option to only show open submodules
- Create a --verbose option which prints status of all submodules and prints working directory

### DIFF
--- a/lib/util/submodule_util.js
+++ b/lib/util/submodule_util.js
@@ -35,6 +35,7 @@
  */
 
 const co      = require("co");
+const colors  = require("colors");
 const NodeGit = require("nodegit");
 const fs      = require("fs-promise");
 const path    = require("path");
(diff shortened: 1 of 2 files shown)


lib/util/submodule_util.js
```diff
@@ -175,7 +175,11 @@ exports.getCurrentSubmoduleShas = co.wrap(function *(repo, submoduleNames) {
     let entry;
     for (let i = 0; i < submoduleNames.length; ++i) {
         entry = index.getByPath(submoduleNames[i]);
-        result.push(entry.id.tostrS());
+        if (entry) {
+            result.push(entry.id.tostrS());
+        } else {
+            result.push("");
```
bpeabody: Let's put something like `colors.red("missing entry")`

bpeabody:
While we're at it, let's change `--summary` to `--list-submodules`; I think `summary` is confusing.

PR #49, opened 2016-05-25
Istanbul Code Coverage Tool

jhedwardyang:
Completes #48 - Generates a full coverage report for code coverage in coverage

- Full mocha integration
- Run via `npm run coverage` or `istanbul cover _mocha`
- Cover a single file the same way as mocha, by passing the filename to above commands, eg. `npm run coverage test/util/branch`
- Generates a fantastic report in coverage/lcov-report/index.html
- Most popular code coverage tool, 4M downloads last month (and currently maintained)

Two points worth noting
- Coverage tool will only show files that are required at least once by a mocha test
- `npm run coverage` will generate coverage, but will fail (return code != 0) if a mocha test fails

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /node_modules/
+/coverage
--- a/package.json
+++ b/package.json
@@ -3,7 +3,8 @@
   "version": "0.2.0",
   "description": "large scale SCM made easy",
   "scripts": {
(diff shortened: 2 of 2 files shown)


bpeabody:
@jhedwardyang sorry I didn't get to this earlier, but it's got conflicts now.

PR #60, opened 2016-06-05
Small clean-up fixes

jhedwardyang:


lib/util/commit.js
```diff
@@ -166,6 +166,6 @@ exports.commit = co.wrap(function *(all, message) {
     // commit anything; we warn the user.
 
     if (null === metaResult) {
-        console.warn("Nothing to commit.");
+        throw new UserError("Nothing to commit.");
```
bpeabody: I don't think we want this to be an error. If a user is writing a script that depends on the exit code, should we end in success or failure in this case. You might check to see what `git commit` does in the same situation.

bpeabody:
Did you really want to close this?

PR #62, opened 2016-07-21
Dturner/pre-receive

novalis:


README.md
```diff
@@ -159,10 +159,29 @@ more information, or one of our test drivers (initially at least
...
+
+- Each sub-repository's SHA in a pushed commit of the meta-repository is
+  either
+
+  - the same as in the meta-repository commit's parent, or
+
+  - directly referenced by the sub-repository's ref named
+    refs/heads/meta-$METASHA, or
+
+  - part of a chain of commits, the top commit of wich meets the
```
abliss: s/wich/which

lib/util/pre_receive_util.js
```diff
@@ -0,0 +1,227 @@
...
+function flatmap(arr, fn) {
+    return arr.map(fn).filter(function (v) { return v; });
+}
+
+class PreReceiveCheck {
+
+    *urlToLocalPath(repo, url) {
+        const config = yield repo.config();
+        const subrepoUrlBase =
+            yield config.getStringBuf("gitmeta.subrepourlbase");
```
abliss: I strongly dislike the smashing together of words with no separator or case-change.

lib/util/pre_receive_util.js
```diff
@@ -0,0 +1,227 @@
...
+        const localPath = path.join(subrepoRootPath, remotePath);
+        if (localPath[0] === "/") {
+            return localPath;
+        } else {
+            return path.normalize(path.join(repo.path(), localPath));
+        }
+    }
+
+    getSyntheticBranchForCommit(commit) {
+        return "refs/git-meta/synthetic/meta-" + commit.id();
```
abliss: This doesn't seem to agree with README.md, which says `/refs/heads/meta-$METASHA`

lib/util/pre_receive_util.js
```diff
@@ -0,0 +1,227 @@
...
+        try {
+            const subrepoCommit =
+                yield submoduleRepo.getReferenceCommit(branch);
+            return subrepoCommit.id().equal(submoduleEntry.id());
+        } catch (e) {
+            return false;
+        }
+    }
+
+    *checkTree(repo, commit, parentTrees, tree) {
```
abliss: This function needs a comment.

lib/util/pre_receive_util.js
```diff
@@ -0,0 +1,227 @@
...
+    *parentLoop(repo, commit, oldSha, handled) {
+        if (commit.id() in handled)  {
+            return true;
+        }
+        handled[commit.id()] = 1;
+        if (oldSha === commit.id().toString()) {
+            return true;
+        }
+
+        const ok = yield GitUtil.readNote(repo, NOTES_REF, commit.id());
```
abliss: Is it worth it to check that the note says "ok"? Could the note ever say something else? Should we even be writing "ok" to the note instead of an empty string?

lib/util/pre_receive_util.js
novalis: My recollection from the git.git source code is that that style is what the APIs required (or at least what was always passed to them).  So I don't want to break anything.

lib/util/git_util.js
```diff
@@ -249,6 +249,28 @@ exports.resolveCommitish = co.wrap(function *(repo, commitish) {
...
+ */
+exports.readNote = co.wrap(function *(repo, notesRef, oid) {
+    assert.instanceOf(repo, NodeGit.Repository);
+    assert.isString(notesRef);
+    assert(typeof(oid) === "string" || oid instanceof NodeGit.Oid);
+
+    try {
+        return yield NodeGit.Note.read(repo, notesRef, oid);
+    }
+    catch (e) {
```
abliss: Doesn't `catch` go on the same line, like `else`?

novalis: In this file, for some reason, no.

novalis:
I just pushed a few changes; they dramatically increase test coverage (and fix some of the bugs I discovered while doing so) .  Note that this code won't work with current node-git; it requires two patches to node-git ( https://github.com/nodegit/nodegit/pull/1080 and https://github.com/nodegit/nodegit/pull/1079 ) , only one of which is merged and neither of which is released.  I tried to make git-meta use the version of nodegit from git, but I couldn't figure it out.

abliss:
LGTM

bpeabody:
I think you need to add `split` to the list of dependencies. After doing that I get the following test failures:

```
  1) pre-receive simplest:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.


  2) pre-receive notes block previous (even bad) history:
     UserError: fail
      at UserError (lib/util/user_error.js:47:26)
      at test/util/pre-receive.js:61:19
      at next (native)
      at onFulfilled (node_modules/co/index.js:65:19)

  3) pre-receive old shas block previous (even bad) history:
     UserError: fail
      at UserError (lib/util/user_error.js:47:26)
      at test/util/pre-receive.js:61:19
      at next (native)
      at onFulfilled (node_modules/co/index.js:65:19)

  4) pre-receive with a submodule, at meta commit:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.


  5) pre-receive with a submodule in a subdir, at meta commit:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.


  6) pre-receive with a submodule in a subdir, from earlier meta-commit:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.


  7) pre-receive with a submodule in a subdir, irrelevant change:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
```

bpeabody:
A few random tabs in the files (I would expect `jshint` to catch these; let me see why it doesn't)

bpeabody:
I'm a little curious about this exception eating in the test driver:

```JavaScript
        try {
            const from = yield x.getReferenceCommit("from");
            old = from.id().toString();
        } catch (e) {
        }
```
Maybe document what's going on there?

bpeabody:
Otherwise, I don't see any obvious reason for the tests to time out. Sorry I have taken so long to look at this; will help out if you need it debugging the test.

novalis:
Are you testing against the patched nodegit?  

The timeout is not good -- there's a thing I don't understand about how promises, generators, etc work, and as a result I've written code which, instead of doing something sensible on exceptions, just times out.

bpeabody:
No, I am not testing against the patched version, let me try that.

novalis:
BTW, now that nodegit has merged my patches (but has not yet done a release), we could change the packages.json to point to some recent SHA of nodegit.  What do you think?

lib/util/pre_receive_util.js
```diff
@@ -0,0 +1,255 @@
...
+        if (success) {
+            yield NodeGit.Note.create(repo, NOTES_REF, newCommit.committer(),
+                                      newCommit.committer(), newAnnotated.id(),
+                                      "ok", 1);
+        }
+        return success;
+    }
+}
+
+PreReceiveCheck.prototype.checkUpdate = co.wrap(
```
bpeabody: There be tabs in this method.

lib/util/repo_ast_util.js
```diff
@@ -652,6 +710,7 @@ exports.cloneRepo = function (original, url) {
             })
         },
         head: head,
+        notes: deepcopy(original.notes),
```
bpeabody: The constructor of `RepoAST` does a copy; you don't need to do it here too.

test/util/shorthand_parser_util.js
```diff
@@ -47,6 +47,7 @@ describe("ShorthandParserUtil", function () {
                 branches: {},
                 remotes: {},
                 index: {},
+                notes:{},
```
bpeabody: I know this is nit-picky, but a space between `:` and `{`, please.

lib/util/pre_receive_util.js
```diff
@@ -0,0 +1,255 @@
...
+    });
+    return yield entryParentTreeEntries.map(
+        function *(entry) {
+            return yield entry.getTree();
+        });
+}
+
+class PreReceiveCheck {
+
+    *urlToLocalPath(repo, url) {
```
novalis: I actually want everything else to be private -- what's the way to do that?

bpeabody: Unfortunately, this is one of the suckier parts of JavaScript. You can get something close to private with ES6 symbols: http://stackoverflow.com/questions/22156326/private-properties-in-javascript-es6-classes, or you can do it the "old school" way with closures as I did with the `SubmoduleRebaser` in `lib/util/rebase.js`.

lib/util/read_repo_ast_util.js
```diff
@@ -36,6 +36,7 @@
  */
 
 const assert   = require("chai").assert;
+const async     = require("generator-async");
```
novalis: co converts a function that returns a generator into a function that returns a promise. I need to convert a function that returns a generator into a sync function.  But I guess I can do it another way.

bpeabody: I think `async.run(foo)` is the equivalent of `co(foo)`. If `async` provides something that `co` doesn't, feel free to keep using `async`.

novalis: oh, right.  The problems is not that co() doesn't work -- it's that errors get swallowed either way, and there were some bugs.  Yay.

lib/util/pre_receive_util.js
```diff
@@ -0,0 +1,255 @@
...
+        const subrepoUrlBase =
+            yield config.getStringBuf("gitmeta.subrepourlbase");
+        const subrepoRootPath =
+            yield config.getStringBuf("gitmeta.subreporootpath");
+        assert(url.startsWith(subrepoUrlBase));
+        const remotePath = url.slice(subrepoUrlBase.length);
+        const localPath = path.join(subrepoRootPath, remotePath);
+        if (localPath[0] === "/") {
+            return localPath;
+        } else {
```
novalis: sorry, does this entail a change here?

test/util/shorthand_parser_util.js
novalis: eslint has an option to enforce this; jshint apparently doesn't. So that's one argument for switching.

lib/util/repo_ast_util.js
novalis: it didn't but it does now.

novalis:
actually I should fix this to match the refs/commits/$commit instead of whatevs/$metacommit.  So don't merge quite yet.

novalis:
hope that doesn't make testing super-unfun.

novalis:
... actually, the tests mocked that.

novalis:
Ok, let me know how this one looks

lib/util/pre_receive_util.js
```diff
@@ -0,0 +1,315 @@
...
+         * parent, it's OK.  Blobs are always OK.  And submodules are
+         * checked with checkSubmodule.
+         * @async
+         * @param {NodeGit.Repostory} repo The meta repository
+         * @param {NodeGit.Commit} commit The meta branch's commit to check
+         * @param {NodeGit.Tree[]} parentTrees the tree objects in parent
+         * commits that correspond to this tree
+         * @param {NodeGit.Tree} tree The tree to check
+         */
+        this[checkTree] = function *(repo, commit, parentTrees, tree) {
```
bpeabody: I believe libgit2 provides a way to look at what's changed in a commit, e.g. `NodeGit.Commit.getDiffWithOptions` that might be simpler (you won't need this recursive tree processing logic) and more efficient. Also, since we really only care about submodules you could probably use `SubmoduleUtil.getSubmoduleChanges`, which will tell you want submodules have been added/changed/removed in a commit. It looks like its documentation is incorrect in that it returns sets, not maps; I'll change.

lib/util/pre_receive_util.js
```diff
@@ -0,0 +1,315 @@
...
+/**
+ * Map a function over an array, returning an array containing only
+ * the elements for which the function returns a truthy value.
+ *
+ * @return {Array}
+ */
+function flatmap(arr, fn) {
+    return arr.map(fn).filter(identity);
+}
+
```
novalis: Maybe, but ramda's "chain" isn't quite the same as flatmap (or maybe I'm abusing the flatmap terminology); it treats a null return as a one-element list instead.

lib/util/pre_receive_util.js
```diff
@@ -0,0 +1,315 @@
...
+        function *(entry) {
+            return yield entry.getTree();
+        });
+}
+
+const urlToLocalPath = Symbol();
+const checkSubmodule = Symbol();
+const checkTree = Symbol();
+const parentLoop = Symbol();
+class PreReceiveCheck {
```
novalis: If I don't do that, then how do I mock methods?

lib/util/pre_receive_util.js
```diff
@@ -0,0 +1,315 @@
...
+         * @param {String} oldSha the previous (known-good) value of this ref
+         * @param {Object} handled the commit ids that have been already
+         * processed.
+         */
+        this[parentLoop] = function*(repo, commit, oldSha, handled) {
+            if (commit.id() in handled)  {
+                return true;
+            }
+            handled[commit.id()] = 1;
+            if (oldSha === commit.id().toString()) {
```
novalis: new refs have an oldSha of 0000000000000000000000000000000000000000

lib/util/pre_receive_util.js
```diff
@@ -0,0 +1,315 @@
...
+        this[checkTree] = function *(repo, commit, parentTrees, tree) {
+            const check = this;
+            const allEntryResults = yield tree.entries().map(function *(entry) {
+                if (entry.isFile()) {
+                    return true;
+                }
+
+                function getEntryByEntryName(parentTree) {
+                    try {
+                        return parentTree.entryByName(entry.name());
```
novalis: I can add an assert for that case, but I bet it fails anyway.

lib/util/pre_receive_util.js
bpeabody: What methods do you need to mock and for what purpose? I don't see us providing a mock implementation of `PreReceive`. Actually, it might make sense to name this util something about its purpose, like `validate_meta_util` rather than its current caller.

lib/util/pre_receive_util.js
bpeabody: Let me rephrase:
1. What case have you encountered that necessitated this try/catch.
2. If it was from a malformed tree (I remember our magic mirror used to write invalid trees), should we explicitly reject it in the same way we would if missing a synthetic meta-branch?

lib/util/pre_receive_util.js
novalis: getSyntheticBranchForCommit for testing.  Although actually I might not need to do that now that I no longer care about the meta commit id, so maybe I can simplify htat.

lib/util/pre_receive_util.js
novalis: I think this is a normal case; a parent tree might just not have this entry (maybe it's newly added).

lib/util/pre_receive_util.js
novalis: Oh wait we still don't know the actual shas of the subrepos, so we can't create the synthetic metabranches, so we do need the class.  But will rename.

novalis:
OK, that submodule diff thingy simplified the code. I wonder why I wrote it the complicated way in the first place.

And I rearranged the commits to be a bit cleaner.  I think I addressed all comments.

Oh, I renamed "synthetic metabranch" to "synthetic branch", since they're now not as related to the meta repo.

lib/util/pre_receive_util.js
```diff
@@ -0,0 +1,315 @@
...
+     * @param {String} newSha the new value of this ref
+     */
+
+    *checkUpdate(repo, oldSha, newSha) {
+        const newAnnotated = yield GitUtil.resolveCommitish(repo, newSha);
+
+        const newCommit = yield repo.getCommit(newAnnotated.id());
+        const success = yield this[parentLoop](repo, newCommit, oldSha, {});
+        if (success) {
+            yield NodeGit.Note.create(repo, NOTES_REF, newCommit.committer(),
```
novalis: I think we'll rarely end up replicating too much work.

lib/util/pre_receive_util.js
bpeabody: Still not sure what you need the class for (not that I have an extremely string opinion on the subject). I think you would need one only if you have state.

bpeabody:
I'll take a fresh look at everything.

PR #65, opened 2016-09-19
Initial implementation of `ls-files`.

bpeabody:
List files from meta-repo and open submodules from current working directory
down.


test/util/list_files.js
```diff
@@ -0,0 +1,102 @@
...
+        },
+        "with a closed sub": {
+            state: "a=S|x=U",
+            expected: ["README.md"],
+        },
+        "open sub": {
+            state: "a=S|x=U:Os",
+            expected: ["README.md", "s/README.md"],
+        },
+        "relative to sub": {
```
novalis: We don't test the "relative to another sub" case.  That is, if you look relative to bar, do you *not* see foo/ files?  Also, these tests should probably check a per-sub filename, so that we know we're actually looking in the sub.

lib/util/list_files.js
```diff
@@ -0,0 +1,109 @@
...
+
+        // If we have a relative path, optimize by skipping submodules whose
+        // contents would be omitted.  Include the contents of this submodule
+        // if the relative path is one of the roots of the submodules path (its
+        // name starts with relative path), or if the relative path is actually
+        // inside the submodule (the submodule name is a prefix of the relative
+        // path).
+
+        if (relativePath &&
+           !(name.startsWith(relativePath) || relativePath.startsWith(name))) {
```
novalis: I'm worried about name="foobar/baz" and relativePath="foo" here -- or does relativePath always have a trailing slash? If so, the preconditions should check this.

test/util/list_files.js
```diff
@@ -0,0 +1,102 @@
...
+            state: "a=S|x=U:Os",
+            relativePath: "s",
+            expected: ["README.md"],
+        },
+        "relative in sub": {
+            state: "a=S|x=U:Os I blam/pow/x.txt=hi",
```
novalis: We should add the following test: state: "a=S|x=U:Os I blam/pow/x.txt=hi", relativePath: "s/morx", expected: [].

lib/util/list_files.js
```diff
@@ -0,0 +1,109 @@
...
+            return [];                                                // RETURN
+        }
+
+        const subRepo = yield SubmoduleUtil.getRepo(repo, name);
+        const files = yield listFilesForRepo(subRepo);
+        return files.map(file => path.join(name, file));
```
novalis: It seems sort of weird to not just filter here for the relative path match.  I can see that we do it below, but why not save space and do it here?

lib/cmd/list_files.js
```diff
@@ -0,0 +1,84 @@
...
+ * Execute the `rebase` command according to the specified `args`.
+ *
+ * @async
+ * @param {Object} args
+ * @param {String} args.commit
+ */
+exports.executeableSubcommand = co.wrap(function *(/* args */) {
+    // TODO: add applicable `git rebase` options.
+
+    const fs   = require("fs-promise");
```
bpeabody: Just to make it lazy, so that we don't have to hit every require any time `git-meta` is invoked.

lib/util/list_files.js
bpeabody: No, I think your worry is justified. I'll add a test case for this. I will make the contract *preclude* a `/` and add one before this check.

bpeabody: (And I missed this item in my last push)

bpeabody:
Ready now.

lib/util/list_files.js
```diff
@@ -0,0 +1,145 @@
...
+
+        // If we have a relative path, optimize by skipping submodules whose
+        // contents would be omitted.  Include the contents of this submodule
+        // if the relative path is one of the roots of the submodules path (its
+        // name starts with relative path), or if the relative path is actually
+        // inside the submodule (the submodule name is a prefix of the relative
+        // path).
+
+        if (relativePath &&
+           !(inRelativePath(name) || relativePath.startsWith(name))) {
```
novalis: I think this still misses the relativePath=foobar and name=foo case.

bpeabody: Indeed it does! I'll fix, but I don't think I can get a test case for it as it's just an optimization.

PR #72, opened 2016-11-10
move node stuff into node directory

bpeabody:
Moving all the node code to its own folder in case we want to add, e.g. Java or C++ code.


jhedwardyang:
LGTM but conflicts due to David's change.

PR #74, opened 2016-11-10
expectation transformation in testMultiRepoManipulator

bpeabody:
Changed `testMultiRepoManipulator` to allow the caller to transformn the map of
expected ASTs immediately prior to checking them against the actual result.
This capability is necessary to test cases where the actual state of the
repository may contain generated (mapped) values where the system cannot adjust
them -- for example, if a branch (or ref) name has the ID of a commit embedded
in it.

### DIFF
--- a/lib/util/repo_ast_test_util.js
+++ b/lib/util/repo_ast_test_util.js
@@ -183,6 +183,12 @@ exports.createMultiRepos = co.wrap(function *(input) {
  *   If it returns `undefined` then it is assumed no mapping is necessary.  The
  *   behvior is undefined if either map contains entries for commits or urls
  *   that already existed in the original map.
+ * - If `options.expectedTransformer` is provided, the map of expected ASTs,
+ *   along with an object containing mapping information will be passed to it
+ *   with the expectation that it will return a (potentially changed) expected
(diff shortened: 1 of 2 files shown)


lib/util/repo_ast_test_util.js
```diff
@@ -193,9 +199,18 @@ exports.createMultiRepos = co.wrap(function *(input) {
...
  * @param {(repoMap, { commitMap, urlMap}) => Promise} manipulator
  * @param {Boolean|undefined}    shouldFail
+ * @param {Object}               [options]
+ * @param {Function}             [options.expectedTransformer]
+ * @param {Object}               options.expectedTransformer.expected
+ * @param {Object}               options.expectedTransformer.mapping
+ * @param {Object}               options.expectedTransformer.mapping.commitMap
+ * @param {Object}               options.expectedTransformer.mapping.urlMap
+ * @param {Object}               options.expectedTransformer.mapping.reverseMap
+ * @param {Object}            options.expectedTransformer.mapping.reverseUrlMap
```
novalis: nit: spacing looks wrong here -- might just be github, but check it.

PR #75, opened 2016-11-10
Some minor hook patches

novalis:
These bugs fixed by the first patch weren't caught by tests because in one case, the buggy method was mocked, and in another, we didn't actually test the hook itself because that would involve spinning up a whole new process, which our framework doesn't really support.

I tested this manually in success and in failure cases, and it seems to work now.

The second patch handles the traditional "foo.git" name for a bare repository.


jhedwardyang:
LGTM.

PR #84, opened 2016-11-20
Rebranding updates

bpeabody:
Fixed most of the commands to have synthetic-meta-ref logic. I include these in one PR because they depended on implementation code that changed and could not be done individually.

### DIFF
--- a/node/lib/cmd/cherrypick.js
+++ b/node/lib/cmd/cherrypick.js
@@ -83,7 +83,8 @@ exports.executeableSubcommand = co.wrap(function *(args) {
     const Status     = require("../util/status");
 
     const repo = yield GitUtil.getCurrentRepo();
-    yield Status.ensureCleanAndConsistent(repo);
+    const status = yield Status.getRepoStatus(repo);
+    Status.ensureCleanAndConsistent(status);
 
     for (let i = 0; i < args.commits.length; ++i) {
         let commitish = args.commits[i];
--- a/node/lib/util/git_util.js
+++ b/node/lib/util/git_util.js
@@ -200,9 +200,7 @@ exports.push = co.wrap(function *(repo, remote, source, target) {
     assert.isString(target);
 
     const execString = `\
-cd ${repo.workdir()}
-git push ${remote} ${source}:${target}
-`;
+git -C '${repo.workdir()}' push ${remote} ${source}:${target}`;
     try {
         yield exec(execString);
         return null;
@@ -299,10 +297,7 @@ exports.fetch = co.wrap(function *(repo, remoteName) {
     assert.instanceOf(repo, NodeGit.Repository);
     assert.isString(remoteName);
 
-    const execString = `\
-cd '${repo.workdir()}'
-git fetch -q '${remoteName}'
-`;
+    const execString = `git -C '${repo.workdir()}' fetch -q '${remoteName}'`;
     try {
         return yield exec(execString);
     }
--- a/node/lib/util/git_util.js
+++ b/node/lib/util/git_util.js
@@ -307,6 +307,48 @@ exports.fetch = co.wrap(function *(repo, remoteName) {
 });
 
 /**
+ * Fetch the specified `sha` from the "origin" repository of the specifieded
+ * `repo`.
+ *
+ * @async
+ * @param {NodeGit.Repository} repo
+ * @param {String} sha
+ */
+exports.fetchSha  = co.wrap(function *(repo, sha) {
+    assert.instanceOf(repo, NodeGit.Repository);
+    assert.isString(sha);
+
+    // First, try to get the commit.  If we succeed, no need to fetch.
+
+    try {
+        yield repo.getCommit(sha);
+        return;                                                       // RETURN
+    }
+    catch (e) {
+    }
+
+    let origin;
+    try {
+        origin = yield repo.getRemote("origin");
+    }
+    catch (e) {
+        throw new UserError(`No origin at ${repo.workdir()}.`);
+    }
+
+    const execString = `\
+git -C ${repo.workdir()} fetch-pack -q '${origin.url()}' ${sha}
+`;
+    try {
+        return yield exec(execString);
+    }
+    catch (e) {
+        throw new UserError(e.message);
+    }
+});
+
+
+
+/**
  * Return a list the shas of commits in the history of the specified `commit`
  * not present in the history of the specified `remote` in the specified
  * `repo`.  Note that this command does not do a *fetch*; the check is made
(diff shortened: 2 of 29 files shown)


node/lib/util/status.js
```diff
@@ -204,8 +204,8 @@ from ${colors.red(GitUtil.shortSha(status.commitSha))}.
                     case RELATION.UNKNOWN:
                         result += `
 Change staged to commit ${colors.yellow(GitUtil.shortSha(status.indexSha))} \
-but cannot verify relation to \
-${colors.yellow(GitUtil.shortSha(status.commitSha))} as the repo is closed.
+but cannot verify relation to previous commit \
```
novalis: This message does not seem actionable to me.  If I saw it, I would not be sure what to do to fix it, nor what the implications of not fixing it would be .

node/lib/util/status.js
```diff
@@ -682,31 +675,12 @@ exports.ensureClean = function (metaStatus) {
 exports.ensureConsistent = function (metaStatus) {
```
novalis: The doc for this method no longer matches what we're trying to achieve.

node/lib/util/write_repo_ast_util.js
```diff
@@ -591,6 +591,14 @@ git -c gc.reflogExpire=0 -c gc.reflogExpireUnreachable=0 \
 
         let index = null;
 
+        // If the base repo has a remote, read its url.
+
+        const remotes = ast.remotes;
+        let originUrl = "";
```
novalis: This should be null not the empty string.

node/test/util/git_util.js
```diff
@@ -109,6 +109,34 @@ describe("GitUtil", function () {
...
+                const x = written.repos.x;
+                const result = yield GitUtil.getOriginUrl(x);
+
+                // If expected URL is not null, find the actual written url.
+
+                let expected = c.e;
+                if (null !== expected) {
+                    for (let key in written.urlMap) {
+                        if (expected === written.urlMap[key]) {
+                            expected = key;
```
novalis: This seems confusing to me -- can you add some comments explaining what this is doing?

node/lib/util/git_util.js
```diff
@@ -128,6 +128,27 @@ exports.isValidRemoteName = co.wrap(function *(repo, name) {
 });
 
 /**
+ * Return the URL for the remote named "origin" in the specified `repo`, or
+ * null if there is no remote named "origin".
```
bpeabody: They mostly end up passing the result to `SubmoduleConfigUtil.resolveSubmoduleUrl`, which will throw a `UserError` only if the submodule URL is relative. I wanted to have this behavior so that everything works in the case that the meta-repo doesn't have an origin, but no relative URLs are used -- frequently in testing.

node/lib/util/git_util.js
```diff
@@ -307,6 +307,48 @@ exports.fetch = co.wrap(function *(repo, remoteName) {
...
+    }
+
+    const execString = `\
+git -C ${repo.workdir()} fetch-pack -q '${origin.url()}' ${sha}
+`;
+    try {
+        return yield exec(execString);
+    }
+    catch (e) {
+        throw new UserError(e.message);
```
bpeabody: I want to lose the stack trace here. We throw`UserError` specifically because we want to show the error to the user; generally, those should not have stack traces.

node/lib/util/submodule_config_util.js
```diff
@@ -68,6 +70,67 @@ const CONFIG_FILE_NAME = "config";
 exports.modulesFileName = ".gitmodules";
 
 /**
+ * Return the result of resolving the specified `relativeUrl` onto the
```
bpeabody: awesome!

node/lib/util/rebase.js
```diff
@@ -135,7 +133,6 @@ function SubmoduleRebaser(submoduleName, repo, branchName, commits) {
 
         if (null === rebase) {
             const subCommit = yield NodeGit.Commit.lookup(repo, commitSha);
-            yield repo.checkoutBranch(branchName);
             yield NodeGit.Reset.reset(repo,
```
bpeabody: ok

bpeabody:
Think I got everything.

node/lib/cmd/cherrypick.js
```diff
@@ -83,7 +83,8 @@ exports.executeableSubcommand = co.wrap(function *(args) {
     const Status     = require("../util/status");
 
     const repo = yield GitUtil.getCurrentRepo();
-    yield Status.ensureCleanAndConsistent(repo);
+    const status = yield Status.getRepoStatus(repo);
+    Status.ensureCleanAndConsistent(status);
```
novalis: [this one too]

novalis:
I just re-pinged on two of them that you missed.  Also please squash commits before merging.

bpeabody:
New
```
=============================== Coverage summary ===============================
Statements   : 96.88% ( 2888/2981 )
Branches     : 91.52% ( 809/884 )
Functions    : 93.31% ( 251/269 )
Lines        : 96.87% ( 2876/2969 )
================================================================================
```

Old
```
=============================== Coverage summary ===============================
Statements   : 97.03% ( 2873/2961 )
Branches     : 92.09% ( 815/885 )
Functions    : 92.91% ( 249/268 )
Lines        : 97.01% ( 2859/2947 )
================================================================================
```

I looked at a few of the differences and didn't see anything too concerning. E.g., I added some assertions like this that aren't being executed; really need something like `isNullOrString`.

```JavaScript
if (null !== repoOriginUrl) {
    assert.isString(repoOriginUrl);
}
```

node/lib/cmd/cherrypick.js
bpeabody: I'm not actually refactoring this part of the code, but I would probably not want choose to factor it that way if I did. It may be too philosophical to discuss in a review, but `RepoStatus` is what I call a "value" type -- it has almost no behavior associated with it; it's job is just to represent the status of a repository. Business logic, such as deciding what "consistent" means is something I put into separate modules.

bpeabody:
Think I got them, squashed commits.

bpeabody:
And one final update to add a TODO for the status message about unknown commits.

node/lib/util/include.js
```diff
@@ -64,46 +64,21 @@ exports.include = co.wrap(function* (repo, url, path) {
...
-    // And check it out.
-
-    yield submoduleRepo.checkoutBranch(branch);
+    const commit = yield submoduleRepo.getBranchCommit(remoteBranch);
 
+    // Set up HEAD and checkout.
 
-    // It looks like you're supposed to call 'addFinalize' to finish setting up
-    // a submodule, but I'm not sure what it does.
+    yield GitUtil.setHeadHard(submoduleRepo, commit);
```
novalis: Should this one also be a checkout?

bpeabody: Since the repo, in this case, was just opened, I don't think the situation you referenced earlier can be a problem.

PR #85, opened 2016-11-20
Rebranding commit

bpeabody:
depends on https://github.com/twosigma/git-meta/pull/84


novalis:
I'm confused.  This PR appears to be built on #84 but the only change outside of #84  is test changes.  That doesn't seem right to me.

bpeabody:
Yes, you're right to be suspicious, but I actually didn't need to make any changes to `commit`. I did change the test cases to reflect the way we expect the tool to be used (detached heads in submodules rather than on branches), but the changes weren't strictly necessary.

PR #86, opened 2016-11-20
Rebranding checkout

bpeabody:
depends on https://github.com/twosigma/git-meta/pull/84

### DIFF
--- a/node/lib/util/checkout.js
+++ b/node/lib/util/checkout.js
@@ -39,102 +39,58 @@ const colors  = require("colors");
 const NodeGit = require("nodegit");
 
 const GitUtil       = require("./git_util");
-const RepoStatus    = require("./repo_status");
 const SubmoduleUtil = require("./submodule_util");
 const UserError     = require("./user_error");
 
 /**
- * Checkout the branch having the specified `branchName` in the specified
- * `metaRepo` having the specified `metaStatus` and all visible sub-repos.  If
- * the specified `create` is "all" then the behavior is undefined if any repo
- * already has a branch named `branchName`.  If `create` is "none" then throw a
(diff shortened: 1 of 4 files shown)


novalis:
It's worth adding a note explaining how this differs from the forthcoming git checkout --recurse-submodules (see https://www.spinics.net/lists/git/msg290281.html et seq).  I think the answer is that it fetches synthetic metarefs if necessary.

node/lib/util/checkout.js
```diff
@@ -39,102 +39,58 @@ const colors  = require("colors");
...
+        const repo = yield SubmoduleUtil.getRepo(metaRepo, name);
+        const sha = shas[name];
+        yield GitUtil.fetchSha(repo, sha);
+        const commit = yield repo.getCommit(sha);
+        yield GitUtil.setHeadHard(repo, commit);
+    }));
 
-    // TODO: I believe there is a bug in nodegit/libgit2 somewhere that results
-    // in a crash 
+    // if 'committish' is a branch,  make it current
```
novalis: Shouldn't it be made current regardless of whether it's a branch or a SHA?  That is, if you want to get to a detached HEAD state, you should be able to.

bpeabody: It will put you in a detached head state if you don't specify a branch. By this comment, I mean that when commitish is also the name of a branch, we'll make it the current branch.

bpeabody:
I'm not sure we need to explicitly reference changes that aren't in Git yet, but I will mention fetching the meta-refs.

bpeabody:
Actually, we're not fetching meta-refs, we're just directly fetching SHAs.

PR #94, opened 2016-11-26
Added `reset` command.

bpeabody:


node/lib/cmd/reset.js
```diff
@@ -0,0 +1,163 @@
...
+    let commit;
+    if (args.commitish) {
+
+        const commitish = args.commitish;
+        const annotated = yield GitUtil.resolveCommitish(repo, commitish);
+
+        if (null === annotated) {
+            throw new UserError(`Could not resolve ${colors.red(commitish)}.`);
+        }
+        commit = yield repo.getCommit(annotated.id());
```
bpeabody: This method does throw (I wish it returned null/error instead), but I think we can assume that the fact that the `AnnotatedCommit` is valid (i.e., we resolve the comitish) implies the existence of a valid commit.

node/lib/cmd/reset.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+    else {
+        commit = yield repo.getHeadCommit();
+    }
+
+    // Perform the reset.
+
+    let type = Reset.TYPE.MIXED;
+
+    if (args.soft) {
```
bpeabody: Yes, I agree.

bpeabody:
Think I got everything.

node/lib/cmd/reset.js
```diff
@@ -0,0 +1,161 @@
...
+        action: "storeConst",
+        constant: true,
+        help: `Do not change the working directory or index.`,
+    });
+
+    parser.addArgument(["--mixed"], {
+        required: false,
+        action: "storeConst",
+        constant: true,
+        help: `Unstage any changes, but to not alter the working dictory.  \
```
novalis: s/to not/do not/

bpeabody: got it, thanks

PR #97, opened 2016-11-28
Fix status to display unknown workdir change.

bpeabody:
Without this fix, if we don't have the commit referenced in the index of a
submodule locally, we don't display anything to indicate there is a change.

### DIFF
--- a/node/lib/util/status.js
+++ b/node/lib/util/status.js
@@ -208,8 +208,8 @@ from ${colors.red(GitUtil.shortSha(status.commitSha))}.
 
                         result += `
 Change staged to commit ${colors.yellow(GitUtil.shortSha(status.indexSha))} \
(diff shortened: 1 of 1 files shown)


node/lib/util/status.js
```diff
@@ -250,6 +250,19 @@ Open repo has unrelated commit \
...
             break;
+        case RELATION.UNKNOWN:
+            // TODO: when we have an appropriate command, such as
+            // `git meta fetch`, recommend using it here instead of
+            // giving this obtuse diagnosis.
+
+            result += `\
+New commit in working directory \
+${colors.yellow(GitUtil.shortSha(status.repoStatus.headCommit))} \
+but cannot verify relation to previous commit \
```
novalis: "but cannot find previous commit ${..} to compare to."

PR #100, opened 2016-11-28
Fix ffwd rebase

bpeabody:


node/lib/util/rebase.js
```diff
@@ -42,6 +42,18 @@ const SubmoduleUtil = require("./submodule_util");
 const UserError     = require("./user_error");
 
 /**
+ * Put the head of the specified `repo` on the specified `commitSha`.
+ */
+const setHead = co.wrap(function *(repo, commitSha) {
+    // TODO: use a more "gentle" strategy that won't stomp on untracked files.
+    // The checkout strategy I was using doesn't handle all situations well,
```
novalis: "was using"?  I'm confused by this comment.

node/lib/util/rebase.js
```diff
@@ -448,6 +459,27 @@ Conflict rebasing the submodule ${colors.red(rebaser.path())}.`;
         yield rebaser.finish();
     }));
 
+    // If this was a fast-forward rebase, we need to set the heads of the
```
bpeabody: Basically, I have to do the equivalent of a `git submodule update` in each affected submodule or they're sitting on the wrong commit at the end of the rebase. In a non-ffwd rebase, I adjust the submodules as I process the commits.

node/lib/util/rebase.js
bpeabody: This code is being called in two places; one of those places was using `checkout` previously.

novalis: I feel like comments ought to explain why the code is the way it is, not how it used to be.

So, "checkout won't work because it doesn't affect nonconflicting staged changes" would work.

node/lib/util/rebase.js
novalis: OK, that makes sense.

PR #101, opened 2016-11-28
Add option to automatically remove directories.

bpeabody:
I grew scared at the number of times I was running "rm -rf *".  Now, if you
run, e.g. `write-repos -o 'a=S'`, the directory `a` will be removed
automatically.

### DIFF
--- a/node/lib/write-repos.js
+++ b/node/lib/write-repos.js
@@ -38,6 +38,8 @@
 const ArgumentParser = require("argparse").ArgumentParser;
 const co             = require("co");
 const mkdirp         = require("mkdirp");
+const path           = require("path");
+const rimraf         = require("rimraf");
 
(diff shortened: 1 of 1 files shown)


node/lib/write-repos.js
```diff
@@ -62,13 +64,28 @@ parser.addArgument(["shorthand"], {
...
 const args = parser.parseArgs();
 
 co(function *() {
     try {
         mkdirp.sync(args.target);
         const shorthand =
                    ShorthandParserUtil.parseMultiRepoShorthand(args.shorthand);
+        if (args.overwrite) {
+            yield Object.keys(shorthand).map(co.wrap(function *(name) {
+                const dir = path.join(args.target, name);
```
novalis: can we add an assert here to make sure "name" is a non-empty string?

bpeabody: The shorthand parser won't parse empty names.

PR #103, opened 2016-11-28
Submodule template

bpeabody:


novalis:
I'm confused here -- isn't this one just https://github.com/twosigma/git-meta/commit/fbd890a8aba1aeaa824ab60e81c71f90d57a1220 + https://github.com/twosigma/git-meta/pull/100 + https://github.com/twosigma/git-meta/pull/101 ?

bpeabody:
I pushed the wrong changes here.

bpeabody:
Should be good now.

node/lib/util/submodule_config_util.js
```diff
@@ -324,9 +325,25 @@ exports.initSubmoduleAndRepo = co.wrap(function *(repoUrl,
...
+    // TODO: fix NodeGit/libgit2 to support the `templatePath` argument.
+
+    const templatePath = path.join(repoPath, ".git", "submodule_template");
+    try {
+        yield fs.stat(templatePath);
+        yield (new Promise(callback => {
+            return ncp(templatePath, result.path(), callback);
+        }));
+    }
+    catch(e) {
```
novalis: Is there a particular error you expect here?  Shouldn't we at least print a warning?

bpeabody: I agree that the way this was written was confusing. See if the new code is more clear.

bpeabody: Well, not just confusing, but also incorrect in that too much code was wrapped by the try/catch.

node/lib/util/submodule_config_util.js
```diff
@@ -324,9 +325,38 @@ exports.initSubmoduleAndRepo = co.wrap(function *(repoUrl,
...
+    // TODO: fix NodeGit/libgit2 to support the `templatePath` argument.
+
+    const templatePath = path.join(repoPath, ".git", "submodule_template");
+
+    // First, check for existence of template directory.
+
+    let hasTemplateDirectory = false;
+
+    try {
+        yield fs.stat(templatePath);
```
novalis: Prefer fs.access().

novalis: (Also, if possible, check that the error is ENOENT and throw otherwise)

bpeabody: got it

bpeabody:
Added documentation of this behavior to the 'open' command.

node/lib/util/submodule_config_util.js
```diff
@@ -324,9 +325,43 @@ exports.initSubmoduleAndRepo = co.wrap(function *(repoUrl,
...
+    let hasTemplateDirectory = false;
+
+    try {
+        yield fs.access(templatePath);
+        hasTemplateDirectory = true;
+    }
+    catch(e) {
+        // no template directory
+
+        const ENOENT = -2;
```
bpeabody: ok, using that module (though in a slightly different manner)

abliss:
Actually, could we get this directory from a config instead? meta.submoduleTemplatePath ?

bpeabody:
Actually, if you don't mind, I'd like to land this for now and then talk some more about the best solution long term. I think it would be reasonable to have a default when there is no configuration anyway. In fact, I think the ideal default would be `./submodule_template` -- users get the configuration automatically when they clone the meta-repo without having to perform any extra steps.

abliss:
The stuff in the submodule templates is generally site-specific. Getting the directory from a (global) config would allow the directory to point to a deployed directory which can be managed centrally. Updates to .git/submodule_template would require messing with existing clones, while messing with ./submodule_template would require rewriting history.

bpeabody:
OK, I can buy that. Let me see if I can wire it up to configuration as @novalis suggested.

bpeabody:
OK, see how this looks now.

abliss:
Hm, one problem with this is that it seems to copy the files *after* initializing the repo. So if you try to put your config options in .git/submodule_template/config , then it overwrites the module's crucial (and specific) configs like core.workTree which are set up in the init process.

bpeabody:
It's the same behavior as `git init`, from `git init --help`:

```
TEMPLATE DIRECTORY
       The template directory contains files and directories that will be copied to the $GIT_DIR after it is created.
```

bpeabody:
I guess that wording could allow for the files being copied after the `.git` directory is created but before it's initialized, but I verified:

```bash
$ git init bar
$ cat bar/.git/description
Unnamed repository; edit this file 'description' to name the repository.
$ mkdir foo
$ echo "hello world" > foo/description
$ git init --template=./foo baz
$ cat baz/.git/description
hello world
```

abliss:
Oops, git config keys are not allowed to contain underscores. submoduleTemplate I guess.

node/lib/util/open.js
```diff
@@ -34,33 +34,34 @@
...
+ * @param {String|null}        repoOriginUrl
+ * @param {NodeGit.Repository} metaRepo
+ * @param {String}             submoduleName
+ * @param {String}             url
+ * @param {String}             commitSha
  * @return {NodeGit.Repository}
  */
 exports.openOnCommit = co.wrap(function *(repoOriginUrl,
-                                          repoPath,
+                                          metaRepo,
```
abliss: Doesn't this require a corresponding change in cmd/open.js where it calls openOnCommit?

bpeabody: Nice catch! Ithought I'd got that. Need to get test drivers for the `cmd` modules. Should be good now.

abliss:
When I try to open something in this branch, after fixing the missed change in cmd/open.js, I always get "no head commit".

bpeabody:
Can you give a repro? When I run, e.g.:
```bash
$ write-repos -o "a=B|x=U"
$ cd x
$ git meta open s
```
it works

abliss:
Nevermind, I had a half-initialized module again.

bpeabody:
I will fix that soon!

node/lib/util/submodule_config_util.js
```diff
@@ -271,34 +273,35 @@ exports.initSubmodule = co.wrap(function *(repoPath, name, url) {
...
- * specified `repoPath`.  Configure the repository for this submodule to have
- * `url` as its remote, unless `url` is relative, in which case resolve `url`
- * against the specified `repoUrl`.  Throw a `UserError` if
- * `null === repoUrl` and `url` is relative.  Return the newly opened
- * repository.  Note that this command does not fetch any refs from the remote
- * for this submodule, and while its repo can be opened it will be empty.
+ * Open the submodule having the specified `name` and `url` for the `metaRepo`.
+ * Configure the repository for this submodule to have `url` as its remote,
+ * unless `url` is relative, in which case resolve `url` against the specified
+ * `repoUrl`.  Throw a `UserError` if `null === repoUrl` and `url` is relative.
```
novalis: Actually asserts, which I think is not a UserError?

abliss:
"It's the same behavior as git init"
Well sort of. It's not the same as "git clone", which does (1) create the .git dir; (2) copy the template stuff; (3) add remotes and fetch.

This version of git-meta seems to (1) create the .git dir, (2) setup the remotes, (3) copy the template stuff, thus overwriting .git/config and breaking the remotes.

node/lib/util/submodule_config_util.js
```diff
@@ -324,9 +327,56 @@ exports.initSubmoduleAndRepo = co.wrap(function *(repoUrl,
...
         workdirPath: workdirPath,
         flags: NO_DOTGIT_DIR | MKPATH | RELATIVE_GITLINK,
     });
+
+    // If there is a `submodule_template` directory in the git folder, copy its
+    // contents into the new `.git` directory for this submodule.
+    //
+    // TODO: fix NodeGit/libgit2 to support the `templatePath` argument.
+
+    // Try to get configuration entry for tempalte path.
```
novalis: spelling on "tempalte"

novalis:
This is approved modulo those two changes; I'll approve it later this evening or tomorrow.

bpeabody:
@abliss I see what you're saying now, but I'm afraid that fixing it will require more severe hackery or resolution of https://github.com/twosigma/git-meta/issues/122. Is it useful for what you need in its current state?

node/lib/util/submodule_config_util.js
bpeabody: I don't think it asserts in that situation. It will assert if `repoUrl` is non-null and non-string.

node/lib/util/submodule_config_util.js
bpeabody: thanks, got it

abliss:
Yep, current state is fine for now.

Issue #104, opened 2016-11-28
`git meta checkout` is very slow and complains about every submodule


abliss:
It says "Submodule X is not clean" for seemingly every submodule.

It also does this kind of crap under strace:
```
[pid 22016] 1480372324.799476 stat("/home/abliss/src/meta/ext/public/twosigma/gitmeta/dist/lib/cmd/version", 0x7ffc955af800) = -1 ENOENT (No such file or directory)
[pid 22016] 1480372324.799532 stat("/home/abliss/src/meta/ext/public/twosigma/gitmeta/dist/lib/cmd/version.js", {st_mode=S_IFREG|0660, st_size=2225, ...}) = 0
[pid 22016] 1480372324.799591 lstat("/home", {st_mode=S_IFDIR|S_ISGID|0755, st_size=4096, ...}) = 0
[pid 22016] 1480372324.799642 lstat("/home/abliss", {st_mode=S_IFDIR|S_ISGID|0755, st_size=12288, ...}) = 0
[pid 22016] 1480372324.799693 lstat("/home/abliss/src", {st_mode=S_IFDIR|S_ISGID|0755, st_size=4096, ...}) = 0
[pid 22016] 1480372324.799740 lstat("/home/abliss/src/meta", {st_mode=S_IFDIR|S_ISGID|0775, st_size=4096, ...}) = 0
[pid 22016] 1480372324.799778 lstat("/home/abliss/src/meta/ext", {st_mode=S_IFDIR|S_ISGID|0775, st_size=4096, ...}) = 0
[pid 22016] 1480372324.799818 lstat("/home/abliss/src/meta/ext/public", {st_mode=S_IFDIR|S_ISGID|0775, st_size=20480, ...}) = 0
[pid 22016] 1480372324.799884 lstat("/home/abliss/src/meta/ext/public/twosigma", {st_mode=S_IFDIR|S_ISGID|0771, st_size=4096, ...}) = 0
[pid 22016] 1480372324.799934 lstat("/home/abliss/src/meta/ext/public/twosigma/gitmeta", {st_mode=S_IFDIR|S_ISGID|0771, st_size=4096, ...}) = 0
[pid 22016] 1480372324.799993 lstat("/home/abliss/src/meta/ext/public/twosigma/gitmeta/dist", {st_mode=S_IFDIR|S_ISGID|0771, st_size=4096, ...}) = 0
[...]

PR #109, opened 2016-11-29
Checkout with dry run

bpeabody:
Should address https://github.com/twosigma/git-meta/issues/104

### DIFF
--- a/node/lib/util/checkout.js
+++ b/node/lib/util/checkout.js
@@ -47,6 +47,8 @@ const UserError     = require("./user_error");
  * `metaRepo`, and update all open submodules to be on the indicated commit,
  * fetching it if necessary.  If `committish` identifies a branch, set that as
  * the current branch.  Throw a `UserError` if `committish` cannot be resolved.
+ * Throw a `UserError` if one of the submodules or the meta-repo cannot be
+ * checked out.
  *
  * @async
  * @param {NodeGit.Repository} repo
@@ -62,7 +64,6 @@ exports.checkout = co.wrap(function *(metaRepo, committish) {
     }
     const commit = yield metaRepo.getCommit(annotated.id());
 
-    yield GitUtil.setHeadHard(metaRepo, commit);
 
     const open = yield SubmoduleUtil.listOpenSubmodules(metaRepo);
     const names = yield SubmoduleUtil.getSubmoduleNamesForCommit(metaRepo,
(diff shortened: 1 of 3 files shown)


node/lib/util/checkout.js
```diff
@@ -62,16 +64,45 @@ exports.checkout = co.wrap(function *(metaRepo, committish) {
...
+    }
+
+    // Try the submodules; store the opened repos and loaded commits for use
+    // in the actual checkout later.
+
+    let subRepos = [];     // will contain a list of repositories for sub-repos
+    let subCommits = [];   // will contain a list of commits to checkout
 
     yield open.map(co.wrap(function *(name) {
         // Open repo but not alive on this commit.
```
novalis: How do we actually want to handle this case?  This is, in effect, checking out a deletion, right?

node/lib/util/checkout.js
```diff
@@ -81,10 +112,51 @@ exports.checkout = co.wrap(function *(metaRepo, committish) {
...
+
+    if (0 !== errors.length) {
+        throw new UserError(errors.join("\n"));
+    }
+
+    /**
+     * Checkout and set as head the specified `commit` in the specified `repo`.
+     */
+    const doCheckout = co.wrap(function *(repo, commit) {
+        yield NodeGit.Checkout.tree(repo, commit, {
```
novalis: I'm confused by running checkout twice.

node/lib/util/checkout.js
```diff
@@ -81,10 +112,51 @@ exports.checkout = co.wrap(function *(metaRepo, committish) {
...
+        });
+        repo.setHeadDetached(commit);
+        yield NodeGit.Checkout.head(repo, {
+            checkoutStrategy: NodeGit.Checkout.STRATEGY.SAFE,
+        });
+    });
+
+    // Now do the actual checkouts.
+
+    yield doCheckout(metaRepo, commit);
```
novalis: A safe checkout strategy on the meta repo is going to refuse to touch modified submodules, and thus is going to fail, right?

node/lib/util/checkout.js
bpeabody: Yes, I guess we should close the repo or else it's going to look like a change to Git. Does that sound reasonable?

node/lib/util/checkout.js
bpeabody: It appears to not have that effect. I'll add a test case to verify.

node/lib/util/checkout.js
bpeabody: The only way I could make it behave like `git checkout` was to run it twice; basically, once to apply to the index and once to the workdir. I'll add a comment to explain.

node/lib/util/checkout.js
bpeabody: Which makes me wonder if we need to do the same thing after reset, rebase, etc. Let's leave this as is for now and discuss. Whatever we decide to do I'll apply uniformly to related functions.

node/lib/util/checkout.js
bpeabody: OK, I added some test cases to verify -- `checkout` does not view any *committed* changes in submodes as being unsafe.

bpeabody:
Think we're good now.

node/lib/util/checkout.js
novalis: That is, nonconflicting committed changes, I assume.

bpeabody: The SAFE option appears to have no effect on the state of the submodules whatsoever.

PR #125, opened 2016-12-02
Perf

bpeabody:
`getRepoStatus` appears to be a performance bottleneck, as it's called on almost all `git-meta` commands.  This PR addresses some of the performance problem by removing assertions from critical paths.

On my laptop, with a meta-repo containing 6,000 repositories, this dropped the time taken to call `getRepoStatus` from ~1,500ms to ~450ms and overall wall time from ~1.9s to ~0.8s.


node/lib/util/write_repo_ast_util.js
```diff
@@ -61,8 +61,14 @@ const TestUtil            = require("../util/test_util");
...
  */
 const doExec = co.wrap(function *(string) {
-    const result = yield exec(string);
-    return result.stdout.split("\n")[0];
+    try {
+        const result = yield exec(string);
+        return result.stdout.split("\n")[0];
+    }
+    catch (e) {
+        console.log(string.split("\n")[1]);
```
novalis: The variable named "string" here should  be called "command".  And maybe just print the whole thing?

node/lib/util/repo_status.js
```diff
@@ -121,146 +121,14 @@ class Submodule {
...
-
+        this.d_indexStatus        = status.indexStatus || null;
+        this.d_indexSha           = status.indexSha || null;
+        this.d_indexShaRelation   = status.indexShaRelation || null;
+        this.d_indexUrl           = status.indexUrl || null;
+        this.d_commitSha          = status.commitSha || null;
+        this.d_commitUrl          = status.commitUrl || null;
+        this.d_workdirShaRelation = status.workdirShaRelation || null;
+        this.d_repoStatus         = status.repoStatus || null;
         Object.freeze(this);
```
novalis: Could we maybe leave the version with assertions under an if (debug) {} or something?

node/lib/util/write_repo_ast_util.js
bpeabody: Got it. The `console.log` in the catch handler was erroneous vestige of debugging; I removed it.

Issue #111, opened 2016-11-30
git meta push currently doesn't work if submodules and meta repository don't have the same remote name


jhedwardyang:
meta remote: upstream
submodules remote: origin

Cannot git meta push -s upstream because submodules do not contain upstream.

bpeabody:
I think that, in fact, `git-meta push` should ignore remotes configured in the submodule and always resolve the submodule URL against the remote being pushed to in the meta-repo -- protecting against potential misconfiguration.

bpeabody:
Fix pending in https://github.com/twosigma/git-meta/pull/135 and related PR https://github.com/twosigma/git-meta/pull/139.

PR #135, opened 2016-12-04
Ignore sub remotes on push

bpeabody:
Note that I need to follow this up with a similar change to `pull`.

### DIFF
--- a/node/lib/cmd/push.js
+++ b/node/lib/cmd/push.js
@@ -50,10 +50,12 @@ exports.description =`
 Push commits from all (visible) sub-repositories and the meta-repository.  By
 default, this command pushes from the current branch to a branch with the
 same name in the 'origin' repository.  The name of the remote, as well as the
-source and target branches may be overridden.  This command will do refuse to
-push the meta-repository unless the remote pushes succeed and the pushed
-branches contain the commits indicated by the submodules in
-the meta-repository.`;
+source and target branches may be overridden.  Push synthetic-meta-refs for
+each submodule with changes indicated by commits being pushed to the meta-repo.
+Don't push to the meta-repo until the submodule pushes finish successfully.  Do
+not consult the remotes configured for the submodules; instead, push the
+synthetic-meta-ref for each submodule to a URL that is derived by resolving
+its configured URL against the URL being pushed to by the meta-repo.`;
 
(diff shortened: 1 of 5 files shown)


jhedwardyang:
I'm not 100% sold on this idea for 2 main points:

1. Having 2 (or more) different servers for repositories seems like a very valid use case. I could want to tie together something on GitLab with something on Gitolite without moving a repository.
2. The behaviour of `git meta push` no longer matches `git push`.

bpeabody:
I agree with the desirability of having multiple remotes (see the [wiki I just wrote](https://github.com/twosigma/git-meta/wiki/Subrepos-Without-Borders) for how this can work), but I don't think you want them to be configured at the submodule level. In fact change in this PR makes it easier to have multiple remotes. Imagine that you have a remote named `origin` and one named `gitlab`. If you ignore the submodule remotes, pushing to `origin` and `gitlab` works naturally: we grab the URL from the selected meta-repo remote and resolve our relative URL against it. If you don't do it this way, you have to make sure all the remotes are synchronized between the submodules and the meta-repo. Furthermore, I don't think it makes any sense to have `git meta push` send some commits to some repos and some to others based on configuration -- most likely such pushes would result in rejection on the back-end.

bpeabody:
@novalis & @jhedwardyang: guys, can you come to an agreement on this today? It's blocking https://github.com/twosigma/git-meta/issues/111 and a related PR: https://github.com/twosigma/git-meta/pull/139.

abliss:
What about ignoring submodule remotes when they are registered with a
relative URL, butty honoring them when absolute? Or offering a config
option to honor them ("never", "absolute", "always")?

bpeabody:
We may learn more about how absolute URLs will be used (or that they just don't work with the overall strategy), but my instinct is that for the meta-repo to make sense, synthetic-meta-refs need to go to the meta-repo-configured location. If we were pushing user-named tags or refs, it would be a different story, but I don't see a valid use-case for users to interfere with the destination of synthetic-meta-refs.

At this moment, I don't want to add configurability until we have a valid use case -- it's just another potential source of problems.

abliss:
SGTM

bpeabody:
@novalis What say ye? Can this go in?

novalis:
Sure.

PR #139, opened 2016-12-06
Ignore remotes configured in submodules during fetch operations.

bpeabody:
Instead, use the `origin` remote from the meta-repo to resolve relative URLs -- this is the same behavior as `git submodule init`. I think we may want to consider a strategy where other remotes are consulted if the fetches fail, or a way for the user to specify a remote command to commands that can trigger a fetch, such as `git meta reset` that would not in the vanilla Git equivalent.

**NOTE** -- depends on https://github.com/twosigma/git-meta/pull/135

### DIFF
--- a/node/lib/util/write_repo_ast_util.js
+++ b/node/lib/util/write_repo_ast_util.js
@@ -515,9 +515,11 @@ exports.writeRAST = co.wrap(function *(ast, path) {
  * @param {Object} repos
  * @param {String} rootDirectory
  * @return {Object}
- * @return {Object} return.repos       map from name to `NodeGit.Repository`
- * @return {Object} return.commitMap   map from new to old commit IDs
- * @return {Object} return.urlMap      map from url to name
+ * @return {Object} return.repos        map from name to `NodeGit.Repository`
+ * @return {Object} return.commitMap    map from new to old commit IDs
+ * @return {Object} return.reverseMap   map from old to new commit IDs
+ * @return {Object} return.urlMap       map from new url to old name
+ * @return {Object} return.reverseUrlMap map from old url to new name
  */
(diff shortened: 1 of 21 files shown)


node/lib/util/write_repo_ast_util.js
```diff
@@ -668,11 +670,16 @@ git -c gc.reflogExpire=0 -c gc.reflogExpireUnreachable=0 \
...
     }
-
+    const reverseUrlMap = {};
+    Object.keys(urlMap).forEach(url => {
+        reverseUrlMap[urlMap[url]] = url;
+    });
     return {
         repos: resultRepos,
         commitMap: commitMaps.newToOld,
+        reverseMap: commitMaps.oldToNew,
```
novalis: reverseCommitMap, for consistency.

node/test/util/submodule_fetcher.js
```diff
@@ -0,0 +1,190 @@
...
+                                                        "a=B:C4-1;Bfoo=4|x=U");
+            const x = written.repos.x;
+            const commit = yield x.getCommit(written.reverseMap["1"]);
+            const fetcher = new SubmoduleFetcher(x, commit);
+            try {
+                yield fetcher.getSubmoduleUrl("s");
+            }
+            catch (e) {
+                assert.instanceOf(e, UserError);
+                    return;                                           // RETURN
```
novalis: nit: indentation

node/lib/util/write_repo_ast_util.js
bpeabody: OK, but I pushed it in a separate commit because I was using that name in other places that had not been changed.

node/test/util/submodule_fetcher.js
bpeabody: got it

Issue #137, opened 2016-12-06
Error: "Remote 'origin' already exists.


novalis:
When attempting a rebase that would (probab;y wrongly, as it happens) open a submodule that had previously been closed.

bpeabody:
@novalis: can you give me repro steps?

PR #143, opened 2016-12-06
Leave modules directory on close.

bpeabody:
Also needed to make `open` support the case where the `.git/modules/${subname}`
directory exists.  The NodeGit routines for openeing submodules fail in this
case, so I had to implement the open operation entirely myself, including
writing the .git link file.

Fixes https://github.com/twosigma/git-meta/issues/137 and other issues such as
when attempting to open a submodule that has a modules directory, such as when
closed using normal git.

### DIFF
--- a/node/lib/util/submodule_config_util.js
+++ b/node/lib/util/submodule_config_util.js
@@ -71,6 +71,34 @@ const CONFIG_FILE_NAME = "config";
 exports.modulesFileName = ".gitmodules";
 
 /**
+ * Return the relative path from the working directory of a submodule having
+ * the specified `subName` to its .git directory.
+ * @param {String} subName
+ */
+exports.computeRelativeGitDir = function (subName) {
+    const depth = subName.split("/").length;
+    let moveUp = "";
+    for (let i = 0; i < depth; ++i) {
+        moveUp = path.join(moveUp, "..");
+    }
+    return path.join(moveUp, ".git", "modules", subName);
+};
+
+/**
+ * Return the relative path from the .git directory of a submodule to its
+ * working directory.
+ * @param {String} subName
+ */
+exports.computeRelativeWorkDir = function (subName) {
+    const depth = subName.split("/").length;
+    let moveUp = path.join("..", "..");
+    for (let i = 0; i < depth; ++i) {
+        moveUp = path.join(moveUp, "..");
+    }
+    return path.join(moveUp, subName);
+};
+
+/**
  * Return the result of resolving the specified `relativeUrl` onto the
  * specified `baseUrl`.
  *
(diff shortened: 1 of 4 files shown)


node/lib/util/submodule_config_util.js
```diff
@@ -334,13 +354,53 @@ exports.initSubmoduleAndRepo = co.wrap(function *(repoUrl,
...
+    const realUrl = exports.resolveSubmoduleUrl(repoUrl, url);
+    let origin = null;
+    try {
+        origin = yield result.getRemote("origin");
+    }
+    catch (e) {
+    }
+    if (null !== origin) {
+        if (realUrl !== origin.url()) {
+            yield NodeGit.Remote.delete(result, "origin");
```
novalis: setUrl?

node/test/util/close.js
```diff
@@ -58,7 +58,7 @@ describe("close", function () {
         const sub = yield NodeGit.Submodule.addSetup(repo,
                                                      baseSubPath,
                                                      "x/y",
-                                                     1);
+                                                      1);
```
novalis: In the github preview, this whitespace change looks wrong.

node/lib/util/submodule_config_util.js
```diff
@@ -71,6 +71,34 @@ const CONFIG_FILE_NAME = "config";
 exports.modulesFileName = ".gitmodules";
 
 /**
+ * Return the relative path from the working directory of a submodule having
+ * the specified `subName` to its .git directory.
+ * @param {String} subName
+ */
+exports.computeRelativeGitDir = function (subName) {
```
novalis: I'm a little concerned about this one.  We had discussed why git has configuration for submodules, and I think I have recalled why: because submodule paths are not unique across history.  So yesterday, my git-remote-hg might have pointed to github.com/felipec/git-remote-hg but today it points to github.com/mnauw/git-remote-hg.  How do we know which one was originally cloned into .git/modules?  That's what the config is for.

So I wonder if it would be better to check the config before manually generating this.

bpeabody: I'm not sure what this concern has to do with `computeRelativeGitDir`, which is just how to point the checked out directory back to the .git directory.

node/lib/util/submodule_config_util.js
bpeabody: yes

node/lib/util/submodule_config_util.js
bpeabody: Why would it change over time? What the config file would tell me? It looks to me that after `git submodule deinit foo` there is no entry in the config file for `foo`.

novalis: It would change over time because the .gitmodules file changes.

But re config file, in that case, I still have no idea what the config file is for, or how the correct mapping from dir to module is supposed to work.  I guess this will do for now, since we're not doing anything truly destructive.

bpeabody: Oh, I do consult the `.gitmodules` file from the `open` command. This utility is just a little lower -level than that and accepts the URL that was previously read from the `.gitmodules`.

PR #146, opened 2016-12-07
Fix read_repo_ast for url-only submodule changes.

bpeabody:
Was previously not correctly recording a change to a submodule if only the URL
was altered.


bpeabody:
Swatted that nit.

PR #149, opened 2016-12-08
Support '-b' option in 'checkout'.

bpeabody:


node/lib/cmd/checkout.js
```diff
@@ -45,9 +45,17 @@ exports.helpText = `Check out a commit in the meta-repository and each
  */
 exports.configureParser = function (parser) {
 
+    parser.addArgument(["-b"], {
+        dest: "new branch name",
```
jhedwardyang: any reason you didn't just name this "branch" (and later args.branch)?

node/lib/cmd/checkout.js
```diff
@@ -45,9 +45,17 @@ exports.helpText = `Check out a commit in the meta-repository and each
...
+        help: `Create a new branch to check out.`,
+        nargs: 1,
+    });
+
     parser.addArgument(["committish"], {
         type: "string",
         help: "commit to check out",
+        defaultValue: null,
+        require: false,
```
jhedwardyang: I think its "required", and also I believe it is invalid (doesn't work) for positional args.

jhedwardyang: You're looking for `nargs: '?'`

bpeabody: yes, thanks!

jhedwardyang:
LGTM.

PR #158, opened 2016-12-13
Work rate limits

bpeabody:


node/lib/util/do_work_queue.js
```diff
@@ -0,0 +1,94 @@
...
+        }
+    });
+
+    // Do the work.  Create an array of `MAX_WORK` items and yield on it.
+
+    let work = [];
+
+    // somewhat-arbitrarily chosen limit on parallel work
+
+    const MAX_WORK = 100;
```
novalis: Why not make this an argument to the function?

node/lib/util/push.js
```diff
@@ -93,24 +94,23 @@ exports.push = co.wrap(function *(repo, remoteName, source, target, force) {
     // First, push the submodules.
 
     let errorMessage = "";
-    const subRepos = yield SubmoduleUtil.getSubmoduleRepos(repo);
     const shas = yield SubmoduleUtil.getSubmoduleShasForBranch(repo, source);
-    const index = yield repo.index();
-    const urls = yield SubmoduleConfigUtil.getSubmodulesFromIndex(repo, index);
```
novalis: isn't the change from index to HEAD a behavior change?

node/lib/util/do_work_queue.js
```diff
@@ -0,0 +1,94 @@
...
+ * at a time.  When `getWork` returns undefined, there is no more work to do.
+ * This method is necessary because calling `yield` on arbitrarily- sized
+ * arrays can lead to problems such as heap exhaustion.  Return an array
+ * containing the result of the work *in order that it was received*, not in
+ * the order it was done.
+ *
+ * @async
+ * @param {() => Promise} getWork
+ */
+exports.doInParallel = co.wrap(function *(getWork) {
```
novalis: If this took an array, it could preallocate the output array instead of having to do the hash->array dance.

node/lib/cmd/open.js
```diff
@@ -78,6 +78,7 @@ exports.configureParser = function (parser) {
 exports.executeableSubcommand = co.wrap(function *(args) {
     const colors = require("colors");
 
+    const DoWorkQueue      = require("../util/do_work_queue");
```
novalis: Do these two commits need to have their order swapped?

node/lib/util/do_work_queue.js
bpeabody: Yeah, but I'd have to pre-allocate an array of functors; doing that up-front allocation (of work) was part of what I wanted to avoid. However, the way I made it does create the "dance" and also complicates the caller's life. What if I make it take a count, and `getWork` takes an index? I think that may give the best of all (or many, at least) worlds.

node/lib/util/push.js
bpeabody: Yes it is. I think as we've evolved our understanding, it's clear (to me) that the way I was doing this was wrong If I'm pushing commit XYZ, I think I should be using the URL configured at commit XYZ rather than something staged in the index. (Maybe I should have mentioned this in the commit message).

node/lib/cmd/open.js
bpeabody: DOH! I'll reorder and push.

bpeabody: Wait, no, unless I'm missing something, 71d4976b7fb52a421a661716899f06a78b10e689 is first commit and adds the `do_work_queue.js` files; 114291f76d3420095e146fc6e18e35bf181d5037 is second and uses them.

bpeabody: (I have to say I'm impressed by the auto-links to those commits)

node/lib/util/do_work_queue.js
novalis: Sure, that's fine.

node/lib/util/push.js
novalis: Apparently, according to the git mailing list discussion I read earlier today, you should be using the one from the config (or the configured remote).  That's to allow old versions with bad URLs to still work.

node/lib/cmd/open.js
novalis: Oh, I misread the log, sorry.

node/lib/util/push.js
bpeabody: I don't think I agree -- which configured remote should we use? I think it's reasonable to allow an option, as we discussed, or other configuration to help guide which remote(s) to fetch from; this will become especially relevant if we opt to go with "true" forking at some point.

Either way, I think that discussion is orthogonal to whether or not to grab the URL from the index or the commit.

bpeabody: I think the strategy espoused by the Git developers may make more sense for submodules as they were intended to be used than as we are using them to build a mono-repo (though the whole story surrounding them is a little confused from the beginning).

node/lib/util/do_work_queue.js
novalis: Actually, you would take a function and an array of inputs, right?  So basically, the interface would be like `map`.

bpeabody: yes, I like that

bpeabody:
OK, I think I've addressed your suggestion; like this much better.

novalis:
You don't seem to have pushed the new commits.

bpeabody:
Maybe this is a bug in Github? I don't see any indication in the conversation view that I've pushed new commits, but if you look at them, you can see that I've changed the method to take an array like we discussed.

PR #167, opened 2017-01-04
Generate repo

bpeabody:
Added utility to generate meta-repos and sub-repos. Improved efficiency of `write_repo_ast_util` and exposed some internals (adding tests for them) used by the generator. Also created a `Stopwatch` class for profiling (finding the ones available in NPM to be lacking).

### DIFF
--- a/node/lib/generate-repo.js
+++ b/node/lib/generate-repo.js
@@ -1,0 +1,457 @@
+#!/usr/bin/env node
+/*
+ * Copyright (c) 2016, Two Sigma Open Source
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of git-meta nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+const ArgumentParser = require("argparse").ArgumentParser;
+const co             = require("co");
+const NodeGit        = require("nodegit");
+const path           = require("path");
+const rimraf         = require("rimraf");
+
+const RepoAST             = require("./util/repo_ast");
+const Stopwatch           = require("./util/stopwatch");
+const SubmoduleUtil       = require("./util/submodule_util");
+const SubmoduleConfigUtil = require("./util/submodule_config_util");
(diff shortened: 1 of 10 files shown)


node/lib/util/stopwatch.js
```diff
@@ -0,0 +1,163 @@
...
+     *
+     * @return {Number}
+     */
+    getCurrent() {
+        assert(0 !== this.d_startCount);
+        return this.getCurrentMs() / 1000.0;
+    }
+
+    /**
+     * the amount of time elapased since `start` was called, in milliseconds.
```
novalis: spelling of elapased.

node/lib/util/stopwatch.js
```diff
@@ -0,0 +1,163 @@
...
+     *
+     * @return {Number}
+     */
+    getCurrentMs() {
+        assert(0 !== this.d_startCount);
+        return (new Date()) - this.d_startTime;
+    }
+
+    /**
+     * total amount of time accumualated in milliseconds
```
novalis: spelling of accumualated

node/lib/util/stopwatch.js
```diff
@@ -0,0 +1,163 @@
...
+const assert = require("chai").assert;
+
+/**
+ * This module provides the `Stopwatch` class.
+ */
+
+/**
+ * `Stopwatch` is a mechanism class providing a way to measure elapsed wall
+ * time.  `Stopwatch` supports *reentrant* code (as is common with 
+ * synchronous code), keeping so that after N calls to `start` it will
```
novalis: s/synchronous/asynchronous/?

But more broadly, you don't use this capability afaict and I don't understand why one would want it (at least, as presently designed).  I could see wanting a timer with subtotals (so that every time you enter a certain function, you add its time to its "account")...

node/lib/util/write_repo_ast_util.js
```diff
@@ -81,181 +64,247 @@ const doExec = co.wrap(function *(command) {
...
+    const builder = yield NodeGit.Treebuilder.create(repo, parentTree);
+
+    const actions = [];  // tree updates to be applied in parallel
 
-        // If no data, make an empty tree
-        if ("" === treeData ) {
-            const builder = yield NodeGit.Treebuilder.create(repo, null);
-            const treeObj = builder.write();
-            return treeObj.tostrS();                                  // RETURN
+    const writeSubtree = co.wrap(function *(filename, entry) {
```
novalis: couldn't this just be called writeTree?

node/lib/util/stopwatch.js
bpeabody: I think I pulled it out of the places I was using it during profiling. E.g.:

```JavaScript
const totalTime = new Stopwatch(true);
yield someArray.map(co.wrap(function *(value) {
    // do some stuff
    //...
    // Now we're going to do the stuff I'm interested in timing.
    totalTime.start();
    // Do interesting stuff that may block.
    totalTime.stop();
    // Do more uninteresting stuff
});

console.log(`Spent ${totalTime.elapsed} doing interesting stuff.`);
```
We're going to end potentially calling `start` many times before matching calls to `stop` are seen. Without this counting capability the behavior of the second call to `start` is undefined (or if it does nothing, the very first call to `stop` will stop the timer). This pattern is pretty common in asynchronous code where there is a high amount of reentrancy within a single thread.

node/lib/util/write_repo_ast_util.js
bpeabody: The outer function is called `writeTree` and this one will never be called on the outermost tree, so I think it makes some sense.

bpeabody:
I think I've got the typos (so far)

node/lib/util/stopwatch.js
novalis: Oh, I see.  Imagine that "interesting stuff" is just sleep(1).  If it happens to get scheduled so that all of the sleeps are sequential, we want to record 100s; if it happens that they're parallel, we want to record 1s (+ epsilon).  And I think this code accomplishes that.

bpeabody: Yes. Do you think I could improve the documentation?

node/lib/util/write_repo_ast_util.js
```diff
@@ -451,18 +535,87 @@ git -C '${repo.workdir()}' checkout-index -a -f
...
 
 /**
+ * Return an array of arrays of commit shas such that the trees of the commits
+ * identified in an array depend only on the commits in previous arrays.  The
+ * tree of one commit depends on another commit (i.e., cannot be created until
+ * that commit exists) if it has a submodule sha referencing that commit.
+ * Until the commit is created, we do not know what its actual sha will be.
+ *
+ * @param {Object} commits map from sha to `RepoAST.Commit`.
+ * @return {[[]] array or arrays of shas
```
novalis: The {[[]] seems wrong.

node/lib/generate-repo.js
```diff
@@ -0,0 +1,457 @@
...
+        for (let i = 0; i < subsToChange; ++i) {
+            const index = randomInt(numSubs);
+            const name = state.submoduleNames[index];
+            if (!(name in subPaths)) {
+                subPaths[name] = true;
+            }
+        }
+    }
+
+    // one commit in five add a sub or always if no subs
```
novalis: s/five/ten/

node/lib/generate-repo.js
```diff
@@ -0,0 +1,457 @@
...
+                makeMetaCommit(state, madeShas, subHeads);
+            }
+            metaCommits += blockSize;
+            const time = new Stopwatch();
+            yield renderBlock(repo, state, madeShas, subHeads);
+            time.stop();
+            doGc(state);
+            console.log(`Writing ${madeShas.length} commits and \
+${subHeads.length} sub changes, took ${time.elapsed} seconds.  Commit \
+rate ${metaCommits / totalTime.elapsed}/S, meta commits ${metaCommits}, \
```
novalis: /S should be /s -- seconds, not siemens.

bpeabody:
Addressed second round of requests.

Issue #95, opened 2016-11-28
git-meta doesn't handle trailing `/` characters in submodule names


bpeabody:
Generally, we should use a filesystem API to normalize submodule names so that, e.g. `../foo` would work too.

abliss:
See also issue #4

Issue #4, opened 2016-04-19
FR: git meta open/close relative dirs


abliss:
If I'm in meta/path/to, and meta/path/to/subrepo is a submodule, I'd like to be able to `git meta close` it using bash's tab-completion help. (This currently works great if my CWD is the root of the repo.)

PR #169, opened 2017-01-05
Add the ability to open/close trees of submodules.

bpeabody:
Now, with `open` and `close`, each "path" argument is interpreted to apply to
all submodules found within this path (this also fixes the trailing "/"
problem).  We will appy this to other meta operations as relevant.

Addresses https://github.com/twosigma/git-meta/issues/4 and https://github.com/twosigma/git-meta/issues/95


bpeabody:
@novalis ping on this, new PR https://github.com/twosigma/git-meta/pull/174 depends on it.

Issue #172, opened 2017-01-10
git meta rebase still leaves staged but uncommited change to meta-repo


jhedwardyang:
Test case. 

`a=B:CX-1;Bmaster=X|x=U:C3-2 a=b;C4-2 s=Sa:X;Bmaster=3;Bfoo=4;Bold=3;Os`

PR #173, opened 2017-01-10
Fix when onto commits have submodule change.

bpeabody:
If the commits being rebased onto have changes to an open submodule, at the end
of a rebase operation we were leaving the HEAD of such submodules on the wrong
commit.  This situation would have left those submodules in a bad state, but it
also caused the rebase commit generation to fail.

Fixes: https://github.com/twosigma/git-meta/issues/172


jhedwardyang:
https://travis-ci.org/jhedwardyang/git-meta/builds/190846223

Issue #175, opened 2017-01-11
`commit` should print status when nothing to commit


bpeabody:
git does this and it helps the user to understand why nothing is being committed.

PR #176, opened 2017-01-17
Status rework

bpeabody:
This change reflects the move towards better support for the mono-repo.  `status` now supports path (and relative path) specification, and `add` is implemented in terms of it. New assumptions provide much more efficient operations with large repositories, and provide the way forward for a path-based `git-meta commit` and `git-meta reset`.

A lot more to do, but wanted to get review started on what's done so far.

addresses https://github.com/twosigma/git-meta/issues/175

### DIFF
--- a/node/lib/cmd/submodule.js
+++ b/node/lib/cmd/submodule.js
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, Two Sigma Open Source
+ * Copyright (c) 2017, Two Sigma Open Source
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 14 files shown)


bpeabody:
To give an idea of the performance benefits, on my laptop (2015 MBP w/ SSD), `git meta status` for an enormous repository (~26k submodules) went from 3s to 400ms.

node/lib/cmd/submodule.js
```diff
@@ -0,0 +1,136 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+"use strict";
+
+const co = require("co");
+
+/**
+ * This module contains methods for pulling.
```
novalis: pulling

node/lib/cmd/submodule.js
```diff
@@ -0,0 +1,136 @@
...
+
+    const repo = yield GitUtil.getCurrentRepo();
+    const workdir = repo.workdir();
+    const cwd = process.cwd();
+    const paths = yield args.path.map(filename => {
+        return GitUtil.resolveRelativePath(workdir, cwd, filename);
+    });
+    const repoStatus = yield Status.getRepoStatus(repo, {
+        paths: paths,
+        showMetaChanges: false,
```
novalis: shouldn't we be getting this from the args?

node/lib/util/commit.js
```diff
@@ -104,6 +106,60 @@ const commitRepo = co.wrap(function *(repo,
...
+        result += Status.printStatusDescriptors(statuses.workdir,
+                                                x => x,
+                                                cwd);
+    }
+    if (0 !== statuses.untracked.length) {
+        result += "\n";
+        result += "Untracked files:\n";
+        result += Status.printUntrackedFiles(statuses.untracked, x => x, cwd);
+    }
+    const prefixed = result.split("\n").map(line => "# " + line).join("\n");
```
novalis: result.replace(/^/mgb, "# ") would be faster and shorter.

node/lib/util/status.js
```diff
@@ -43,235 +53,411 @@ const SubmoduleUtil       = require("../util/submodule_util");
...
+class StatusDescriptor {
+    /**
+     * @param {RepoStatus.FILESTATUS} status
+     * @param {String}                path
+     * @param {String}                detail
+     */
+    constructor(status, path, detail) {
+        this.status = status;
+        this.path = path;
+        this.detail = detail;
```
novalis: Should we check these types?

node/lib/util/status.js
```diff
@@ -43,235 +53,411 @@ const SubmoduleUtil       = require("../util/submodule_util");
...
+ * @param {Function}            color
+ * @return {String}
+ */
+exports.printStatusDescriptors = function (statuses, color, cwd) {
+    assert.isArray(statuses);
+    assert.isFunction(color);
+
+    const sorted = exports.sortDescriptorsByPath(statuses);
+    const lines = sorted.map(status => "\t" + status.print(color, cwd));
+    if (0 !== lines.length) {
```
novalis: you could move this check to the beginning of the function -- if (statuses.length == 0) { return ""; }

node/lib/util/status.js
```diff
@@ -526,40 +788,80 @@ exports.getRepoStatus = co.wrap(function *(repo) {
...
+                 yield SubmoduleConfigUtil.getSubmodulesFromIndex(repo, index);
+        const indexNames = Object.keys(indexUrls);
+        const headUrls =
+           yield SubmoduleConfigUtil.getSubmodulesFromCommit(repo, headCommit);
 
-        const commitTree = yield headCommit.getTree();
 
-        // Make a list of all subs that exist in either the head commit or the
-        // index.
+        // No paths specified, so we'll do all submodules, restricint to open
```
novalis: spelling: restrictint

node/test/util/commit.js
```diff
@@ -59,131 +61,227 @@ const committer = co.wrap(function *(doAll, message, repos) {
...
+
+        const cases = {
+            "basic form": {
+                status: new RepoStatus({
+                    currentBranchName: "master",
+                    staged: {
+                        "foo": FILESTATUS.ADDED,
+                    },
+                }),
+                match: /Changes to be committed/,
```
novalis: I would actually like these to just describe the complete output.  match is error-prone in that it only checks for some deviations

node/lib/util/status.js
bpeabody: I didn't put a check in for these because I've found that the assertions can be prohibitively expensive, and I expect that there may be a great number of these created. For example, the `assert.isObject` at the beginning of the constructor for `RepoAST.Submodule` cost 500ms when invoked 26k times.

I wish JS (and Java) had macros.

I guess I could put a comment explaining why they aren't there, at least.

novalis: Eventually we'll switch to typescript or something and then we won't have to worry about it.

node/lib/cmd/submodule.js
bpeabody: This flag controls whether or not changes to the meta-repo workdir and index are loaded. Changes to the meta-repo workdir and index cannot affect the output of this command.

node/lib/util/status.js
bpeabody: Yeah, that would be great.

node/lib/util/status.js
bpeabody: nice catch

bpeabody:
Think I've got everything so far.

node/lib/util/add.js
```diff
@@ -0,0 +1,85 @@
...
+    assert.isBoolean(stageMetaChanges);
+
+    const repoStatus = yield Status.getRepoStatus(repo, {
+        showMetaChanges: stageMetaChanges,
+        includeClosedSubmodules: false,
+        paths: paths,
+        showAllUntracked: true,
+    });
+
+
```
novalis: nit: extra blank line

PR #180, opened 2017-01-22
Find meta

bpeabody:
Added a command that allows you to find the meta-repo commit that introduced a specific commit in a submodule.


bpeabody:
Re. performance: it takes 45 seconds to the entire history in my test repo containing 260,000 commits on my laptop.

node/lib/util/log_util.js
```diff
@@ -0,0 +1,165 @@
...
+ */
+
+const assert  = require("chai").assert;
+const co      = require("co");
+const NodeGit = require("nodegit");
+
+const SubmoduleFetcher = require("./submodule_fetcher");
+
+/**
+ * Return first ancestor of the specified `metaCommit` in the specified
```
novalis: s/first/most recent/

node/lib/util/log_util.js
```diff
@@ -0,0 +1,165 @@
...
+                                       NodeGit.Oid.fromString(subShaForCommit),
+                                       subCommit.id())) !== 0;
+            }
+        }
+        existsInSha.set(sha, result);
+        return result;
+    });
+
+    while (0 !== toCheck.length) {
+        const commit = toCheck.shift();
```
novalis: Irritatingly, v8 arrays aren't deques; this is an O(n) operation.  https://github.com/petkaantonov/deque seems to be the sane library for this (the one on npm called "deque" uses linked lists, bizarrely)

node/lib/util/log_util.js
```diff
@@ -0,0 +1,165 @@
...
+        const sha = commit.id().tostrS();
+        // Bail if we've already checked this one.
+
+        if (checked.has(sha)) {
+            continue;                                               // CONTINUE
+        }
+
+        const MAX_PARENTS = 500000;  // arbitrary limit
+        const parents = yield commit.getParents(MAX_PARENTS);
```
novalis: commit.getParents(commit.parentcount()) ?

node/lib/util/log_util.js
```diff
@@ -0,0 +1,165 @@
...
+
+        const existsHere = yield doesExistInCommit(commit);
+
+        if (null !== existsHere) {
+            // Stage parents to be checked unless the repo doesn't exist here.
+
+            toCheck = toCheck.concat(parents);
```
novalis: This is also O(n).  Why not just push?

node/lib/util/log_util.js
```diff
@@ -0,0 +1,165 @@
...
+        if (null === existsHere || false === existsHere) {
+            continue;                                               // CONTINUE
+        }
+
+        // If the commit isn't referenced in any parent, this one is the
+        // winner.
+
+        let notInParents = true;
+        for (let i = 0; notInParents && i < parents.length; ++i) {
+            notInParents = true !== (yield doesExistInCommit(parents[i]));
```
novalis: Instead of true !==, what about just ! ?

node/lib/util/log_util.js
bpeabody: I think a `vector` is a better choice than a `deque` here. The list will generally be short (a little larger than the maximum number of parents for a single commit) and `vector` is usually faster than `deque` until you get to containers around word size, the O(N) aspect being dwarfed by other factors (contiguous layout of vectors, CPUs heavily optimized to work with them).

novalis: The deque I linked to is backed by an array, so the perf should be quite good.  And in a very-branching repo, your queue size can get large.

bpeabody:
I think this is ready.

node/lib/cmd/submodule.js
```diff
@@ -135,3 +150,126 @@ exports.executeableSubcommand = co.wrap(function *(args) {
...
+    // referring to no submodule, or by referring to more than one.
+
+    const relPath = yield GitUtil.resolveRelativePath(workdir,
+                                                      process.cwd(),
+                                                      path);
+    const resolvedPaths = yield SubmoduleUtil.resolvePaths(workdir,
+                                                           [relPath],
+                                                           subNames,
+                                                           openSubNames);
+    const paths = Object.keys(resolvedPaths);
```
jhedwardyang: If a submodule has since been removed (after metaCommittish, but before HEAD), is find-meta unsupported? Even if metaCommittish is set?

node/lib/cmd/submodule.js
```diff
@@ -135,3 +150,126 @@ exports.executeableSubcommand = co.wrap(function *(args) {
...
+        subRepo = yield Open.openOnCommit(fetcher, subName, subSha);
+    }
+    else {
+        subRepo = yield SubmoduleUtil.getRepo(repo, subName);
+    }
+
+    // Now we resolve the committishes.
+
+    const metaAnnotated = yield GitUtil.resolveCommitish(repo, metaCommittish);
+    if (null === metaAnnotated) {
```
jhedwardyang: We could probably fail faster than this! Especially since it'll open the sub-repo as a side effect.

node/lib/cmd/submodule.js
bpeabody: If the submodule is gone on HEAD, we won't be able to open it to resolve `subCommittish`; we'll fail immediately. We could, in theory, look at the submodules in `metaCommittish`, but while I'm OK with opening a submodule to make this work, I don't want to change the state of the repo otherwise.

(not sure if this answers your question)

node/lib/cmd/submodule.js
bpeabody: Very good point; I've changed this.

bpeabody:
@jhedwardyang, @novalis: ping

jhedwardyang:
LGTM. 

I think a better way of achieving this would be to upstream a patch to git log -p --submodule=<format>.

`git log -p --submodule=log path/to/submodule` is just missing a commit hash to achieve this functionality.

bpeabody:
Created issue: https://github.com/twosigma/git-meta/issues/186

Issue #188, opened 2017-01-30
add `--amend` flag to `git meta commit`


zls88:
I frequently find myself rebasing my meta repository just to squash commits together. An `--amend` flag to match the one on `git commit` would make like easier.

bpeabody:
Yes. There are a few interesting implementation details to implement this, but it's in progress; hope to have something this week.

PR #189, opened 2017-02-06
Amend

bpeabody:
This change adds support for `git meta commit --amend`. It will not run unless the commits to be amended for the meta-repo and all submodules match (author and commit message). In addition to this check, the command involves several non-obvious complications:

- Computing the new effective diff between the current state of the repo and HEAD^ is non-trivial, particularly in the "-a" case.
- We need the effective diff to show an accurate modified files list to users in the message prompt,
- to know which files need to be staged in the "-a" case,
- and to detect potentially empty commits in submodules. In these cases we need to remove the old commit instead of amending it.
- Also, we need to stage and create non-amend commits in previously un-modified submodules with changes.

Addresses: https://github.com/twosigma/git-meta/issues/188


bpeabody:
Ping: this has been open for five days and addresses a frequent user request. I will get a reviewer from our node community if we don't have time to review it ourselves.

node/lib/cmd/commit.js
```diff
@@ -130,10 +141,120 @@ exports.executeableSubcommand = co.wrap(function *(args) {
...
+    // any.
+
+    let oldSubs = {};
+    const parent = yield GitUtil.getParentCommit(repo, head);
+    if (null !== parent) {
+        oldSubs = yield SubmoduleUtil.getSubmodulesForCommit(repo, parent);
+    }
+
+    // Check to see if the repo is in a valid state to be amended.
+
```
bpeabody: that one I meant to have

node/lib/util/repo_status.js
```diff
@@ -273,6 +273,40 @@ class Submodule {
...
+     * @return {Submodule}
+     */
+    copy(args) {
+        if (undefined === args) {
+            args = {};
+        }
+        else {
+            assert.isObject(args);
+        }
+        return new Submodule({
```
bpeabody: I don't think there's a clean way to do a mapping.

node/lib/util/shorthand_parser_util.js
```diff
@@ -884,6 +884,9 @@ function parseOverrides(shorthand, begin, end, delimiter) {
         workdir: workdir,
         openSubmodules: openSubmodules,
     };
+    if (undefined === head && undefined !== currentBranchName) {
+        head = branches[currentBranchName];
```
bpeabody: In detached head state `undefined === currentBranchName`.

bpeabody:
Think I got everything (so far) and rebased off of `master`.

doc/user-guide.md
```diff
@@ -371,6 +371,89 @@ $ git add my-sub-repo
 $ git commit -m "Added a new file"
 ```
 
+#### Amending Changes
+
+The `git commit --amend` command is used to adjust the most recent commit.  As
```
jhedwardyang: `git meta commit --amend`

doc/user-guide.md
```diff
@@ -371,6 +371,89 @@ $ git add my-sub-repo
...
+$ echo >> sub-repo-b/README.md
+$ cd sub-repo-a
+$ git commit --amend -m "updated READMEs"
+$ cd ../sub-repo-b
+$ git commit -a -m "updated READMEs"
+$ cd ..
+$ git commit -a -m "updated READMEs"
+```
+
+If you undo a change, `git meta commit --amend` will remove the unneeded
```
jhedwardyang: I recognize this is because we want a mono-repo behaviour. However, it is a little odd that I am removing a commit when I "undo" a committed change.  I would expect it to look at HEAD in the submodule.

jhedwardyang: This also creates an interesting situation, where if I want to undo a commit in a submodule, I have to type the following:
`git revert -n HEAD`

Rather than
` git checkout HEAD~`

node/lib/cmd/commit.js
```diff
@@ -118,7 +130,8 @@ exports.executeableSubcommand = co.wrap(function *(args) {
     // this time.
 
     if (repoStatus.areUncommittableSubmodules()) {
-        process.stderr.write(Status.printRepoStatus(repoStatus, relCwd));
```
jhedwardyang: Any particular reason for `process.stderr.write` here, but `console.error` below?

jhedwardyang:
The code itself makes sense. Its a little hard to read the review because of the Status -> StatusUtil change. 

My only concern is with the amend behaviour being odd when HEAD is changed in the submodules.

node/lib/cmd/commit.js
```diff
@@ -144,9 +158,118 @@ ${colors.yellow("Please stage changes in new submodules before committing.")}\
...
+    // Update 'status' to reflect changes to be applied with amend.
+
+    status = amendChanges.status;
+
+    // If no message, use editor.
+
+    if (null === args.message) {
+        const headSig = head.author();
+        const time = headSig.when();
+        const date = new Date(time.time() * 1000);
```
jhedwardyang: `new Date()` should be sufficient

bpeabody: We don't want  a new date with the current time, we want one with the original time.

doc/user-guide.md
bpeabody: right

doc/user-guide.md
bpeabody: When I say "undo", I mean you manually revert the commit such that the result of an amend operation would be to create an empty commit in the submodule. This situation is admittedly very rare. Rather than make an empty commit, I point HEAD of the submodule to HEAD^.

If you attempt to do this with `git commit --amend`, it will refuse and give the following complaint:

```
No changes
You asked to amend the most recent commit, but doing so would make
it empty. You can repeat your command with --allow-empty, or you can
remove the commit entirely with "git reset HEAD^".
```

My feeling is that the natural thing to do in keeping with the mono-repo module is to follow Git's advice for removing the commit entirely.

I don't think this has to anything to do with `git revert` or `git checkout HEAD^`; we're talking about the (unusual) situation where the developer manually undid the change.

node/lib/cmd/commit.js
bpeabody: `console.error` is shorter but appends a newline that would be redundant in this case.

bpeabody:
Fixed the typo you found and hopefully answered your questions.

node/lib/cmd/commit.js
jhedwardyang: I think I misread this. We need to reset the committer date right?

bpeabody: This is just the date we show in the message in the editor. The actual date saved is done by the `NodeGit.Commit.amend` method, which uses the original date that we show here.

doc/user-guide.md
jhedwardyang: In fact, in the git meta workflow, manually undoing a submodule commit tends to be fairly common (due to the nature of `git meta commit -a`).

bpeabody: Then it's especially good that we're addressing this case :) Though if users are frequently inadvertently committing changes to submodules, we might want to make it easier to undo them. Also, I'll be adding a path-based commit soon that should help.

Issue #191, opened 2017-02-07
push to delete branch (`git meta push <remote> :<target>`) is broken


bpeabody:
```
TypeError: Cannot read property 'id' of null
```

PR #192, opened 2017-02-07
Fix push delete syntax.

bpeabody:
Addresses https://github.com/twosigma/git-meta/issues/191

### DIFF
--- a/node/lib/cmd/push.js
+++ b/node/lib/cmd/push.js
@@ -109,10 +109,24 @@ exports.executeableSubcommand = co.wrap(function *(args) {
 
     yield strRefspecs.map(co.wrap(function *(strRefspec) {
         const refspec = GitUtil.parseRefspec(strRefspec);
-        yield push.push(repo,
-            args.repository,
-            refspec.src,
-            refspec.dst,
-            args.force || refspec.force || false);
+        const force = args.force || refspec.force || false;
+
+        // If 'src' is empty, this is a deletion.  Do not use the normal meta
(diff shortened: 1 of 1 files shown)


node/lib/cmd/push.js
```diff
@@ -109,10 +109,24 @@ exports.executeableSubcommand = co.wrap(function *(args) {
...
+
+        if ("" !== refspec.src) {
+            yield push.push(repo,
+                            args.repository,
+                            refspec.src,
+                            refspec.dst,
+                            force);
+        }
+        else {
+            yield GitUtil.push(repo,
```
jhedwardyang: I was hoping we could use one code path for deleting refs. If you want, I could take a stab at this.
Also, we should add a unit test for this.

bpeabody: The other code-path does call into `GitUtil.push`, so we are using one code path for the important parts. Just because the UI for Git chose to meld two logically separate operations (updating a branch and deleting a branch) doesn't necessarily mean we need to architect our implementation that way.

That said, I don't like the fact that what I'm doing here is in the `cmd` and doesn't have a test driver. If you'd like to move some of this logic into `push.js` and have a plan to make it more elegant (without adding extra branches to the code concerned with pushing updates), I would be happy for you to take a stab at it.

bpeabody:
Ping

jhedwardyang:
Won't be able to work on this until next weekend. Approving for now and will come back to move the logic

PR #193, opened 2017-02-07
Refactor submodule name resolution from open/close

bpeabody:
Also fix a minor bug where an invalid path would trigger an erroneous error
message.


bpeabody:
also addresses https://github.com/twosigma/git-meta/issues/194

Issue #196, opened 2017-02-09
work around crash accessing config in parallel from `open`


bpeabody:
see: https://github.com/twosigma/git-meta/issues/195

Basically, we should just read the config value for 'templatePath' once, before opening any submodules. This "workaround" is a reasonable optimization anyway. When opening hundreds of submodules it probably doesn't make sense to keep reprising the config file.

PR #197, opened 2017-02-09
Read template path outside of open/submodule init.

bpeabody:
Previously, we were reading the meta-repo's config file for its submodule
template path deep in `SubmoduleConfigUtil`.  This method was inefficient, and
tripped a bug in libgit2 (or nodegit) where reading the config file in parallel
seems to cause crashing.

(Also addressed a small bug formatting the status of a closed but dirty repo)

Addresses: https://github.com/twosigma/git-meta/issues/196


bpeabody:
Ping: this fixes an ongoing issue with `git meta open` that is affecting users.

PR #198, opened 2017-02-09
Fix type error on open

bpeabody:
Depends on https://github.com/twosigma/git-meta/issues/196 (could go in separately, but would require a fix up).

### DIFF
--- a/node/lib/util/submodule_config_util.js
+++ b/node/lib/util/submodule_config_util.js
@@ -407,7 +407,7 @@ exports.initSubmoduleAndRepo = co.wrap(function *(repoUrl,
     }
     if (null !== origin) {
         if (realUrl !== origin.url()) {
-            yield NodeGit.Remote.setUrl(result, "origin", realUrl);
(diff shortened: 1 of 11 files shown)


bpeabody:
Ping: this fixes an ongoing problem affecting users.

node/lib/util/submodule_config_util.js
```diff
@@ -407,7 +407,7 @@ exports.initSubmoduleAndRepo = co.wrap(function *(repoUrl,
     }
     if (null !== origin) {
         if (realUrl !== origin.url()) {
-            yield NodeGit.Remote.setUrl(result, "origin", realUrl);
+            NodeGit.Remote.setUrl(result, "origin", realUrl);
```
jhedwardyang: Sigh

bpeabody: We didn't find it earlier because users have to get their submodule in a weird state (probably, e.g., from https://github.com/twosigma/git-meta/issues/196) to hit this case.

PR #200, opened 2017-02-11
Better status

bpeabody:
This change includes two big improvements:

1. Factored out general `git meta status` and that used to calculate "amend" status to use the same code path -- abandoning the `Nodegit.Repository.getStatusExt` API, that, it turns out, is insufficient to calculate anything other than the most basic status situation -- workdir to index and index to HEAD.

2. Fixed status calculation in the `-a` case to report correct accumulated status, e.g., when the workdir reverses an index change.

Generally, these changes will make it easy to implement `git meta commit` with paths.

Depends on: https://github.com/twosigma/git-meta/pull/189


node/lib/util/diff_util.js
```diff
@@ -129,41 +129,38 @@ exports.getRepoStatus = co.wrap(function *(repo,
...
+        const indexToTreeDiff =
+            yield NodeGit.Diff.treeToIndex(repo, tree, null, options);
+        const indexToTreeStatus = readDiff(indexToTreeDiff);
         return {
-            staged: indexStatus,
+            staged: indexToTreeStatus,
             workdir: workdirToIndexStatus,
         };
     }
-    const workdirDiff =
```
novalis: git diff usually shows (index-to-tree, worktree-to-index).  So I'm confused about what we're showing here.

bpeabody: The alternate situation is used to derive the diff used when, for example, the user passes `-a` to `commit`, e.g.:

```
$ touch foo # new file
$ git add foo
$ rm foo
$ git status
On branch master
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        new file:   foo

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        deleted:    foo
```

Yet:

```
$ git commit -a
On branch master
nothing to commit, working tree clean
```

Similarly if a file been changed to one state in the index, and back to the original (HEAD) state in the workdir.

novalis: Oh, I see. IIRC, GIt handles this with a temporary index which is just used during the commit.  But I guess this way works too.

node/lib/util/diff_util.js
```diff
@@ -53,6 +53,7 @@ exports.convertDeltaFlag = function (flag) {
         case DELTA.CONFLICTED: return FILESTATUS.CONFLICTED;
         case DELTA.RENAMED: return FILESTATUS.RENAMED;
         case DELTA.TYPECHANGE: return FILESTATUS.TYPECHANGED;
+        case DELTA.UNTRACKED: return FILESTATUS.ADDED;
```
novalis: wait, untracked -> added?  This needs explanation.

bpeabody: Clarified with comment.

bpeabody:
Added clarification re. treatment of `UNTRACKED`. Also made minor cleanup of comment and formatting at end of `diff_util.getRepoStatus`.

Issue #215, opened 2017-03-06
git meta rebase --continue fails with "Conflict rebasing module" when there is no conflict


bpeabody:
There is one commit to rebase on each submodule. In this iteration there was an actual conflict on one submodule. After fixing this conflict git meta rebase --continue is not able to proceed.

bpeabody:
I can recreate this (manually, working on test case). Use `write-repos` with:

```
a=B:Ca-1;Cb-1 a=q;Cc-1;Ba=a;Bb=b;Bc=c|x=S:C2-1 s=Sa:1,t=Sa:1;C3-2 s=Sa:a,t=Sa:a;C4-2 s=Sa:b,t=Sa:c;Bmaster=3;Bfoo=4;Os;Ot
```

then, in repo `x` invoke `git meta rebase foo`, fix the conflict in `s/a`, and try to do `git meta rebase --continue`.

bpeabody:
... it complains about a conflict with `t`. Interestingly, a subsequent `git meta rebase` results in complaints of conflicts in `s` and `t`.

PR #223, opened 2017-03-12
Fix rebase continue with untouched submodules.

bpeabody:
Return for submodules that weren't being rebased made them look like they were
in conflict.  Additionally, I changed it to stage commits that might have been
made in such a repo.  Added tests for both cases.

Addresses: https://github.com/twosigma/git-meta/issues/215


bpeabody:
@novalis ping

Issue #234, opened 2017-03-30
git meta commit --amend incorrectly displays "<submodule> has new commits"


shaoster:
Repro:
- Create a new meta repo.
- open a submodule.
- create a new commit in the submodule.
- git meta commit -a -m "foo"
- soft reset back to baseline.
- git meta commit -a --amend -m "bar"

git meta status reports the correct issue:
"(submodule, on old commit)"

bpeabody:
After discussion with @shaoster , I believe this issue is caused by a combination of confusion/user error.

bpeabody:
Completely redid this.

PR #235, opened 2017-03-31
Accurate error message on amend failure.

bpeabody:
Instead of always saying "new commits", display the actual commit relation.

Addresses: https://github.com/twosigma/git-meta/issues/234


bpeabody:
@novalis @jhedwardyang ping

PR #242, opened 2017-04-14
Fix close.

bpeabody:
Was broken after earlier merge bug.


bpeabody:
Good point. Moved existing `util/close.js` into a different module, made a new `util/close_util.js` module with this logic and added a test driver.

Issue #238, opened 2017-04-06
Allow separate commit messages for each submodule in `commit`.


bpeabody:
E.g., would be split out like:

```
# <*> enter meta-repo message above this line; delete to commit only submodules
# On branch master.
# -----------------------------------------------------------------------------

# <s> enter message for 's' above this line; delete this line to skip committing 's'
# Changes to be committed:
# 	new file:     a
# -----------------------------------------------------------------------------

# <t> enter message for 't' above this line; delete this line to skip committing 't'
# Changes to be committed:
# 	modified:     README.md
#
# Please enter the commit message(s) for your changes.  The message for a
# repo will be composed of all lines not beginning with '#' that come before
# its tag, but after any other tag (or the beginning of the file).  Tags are
# lines beginning with '# <sub-repo-name>', or '# <*>' for the meta-repo.
# If the tag for a repo is removed, no commit will be generated for that repo.
# If you do not provide a commit message for a sub-repo, the commit
# message for the meta-repo will be used.
```

Would also apply to `amend`.

bpeabody:
Here is the prompt syntax I'm going with (there being changes to submodules `s` and `t` in this example):
```

# <*> enter message for the meta-repo above this line; delete to skip
# On branch master.
# -----------------------------------------------------------------------------

# <s> enter message for 's' above this line; delete to skip
# Changes to be committed:
# 	new file:     a
# -----------------------------------------------------------------------------

# <t> enter message for 't' above this line; delete to skip
# Changes to be committed:
# 	modified:     README.md
#
# Please enter the commit message(s) for your changes.  The message for a
# repo will be composed of all lines not beginning with '#' that come before
# its tag, but after any other tag (or the beginning of the file).  Tags are
# lines beginning with '# <sub-repo-name>', or '# <*>' for the meta-repo.
# If the tag for a repo is removed, no commit will be generated for that repo.
# A sub-repo with an empty commit message will inherit the commit message of
# the meta-repo, if it has one.  Commits will not be created for repos with
# empty commit messages.
```

novalis:
"delete to skip" on the meta repository is a little unclear -- what does it mean?

Also the following seems self-contradictory:

"A sub-repo with an empty commit message will inherit the commit message of the meta-repo, if it has one.  Commits will not be created for repos with empty commit messages."

bpeabody:
Re. the first item, how about: "delete this line to skip committing this repo"

Re. the second, how does this sound:
[...]

PR #245, opened 2017-04-18
Commit split

bpeabody:
If we like this, I'll make it work for `amend` too. Would welcome feedback on the syntax.

### DIFF
--- a/node/lib/util/commit.js
+++ b/node/lib/util/commit.js
@@ -50,6 +50,7 @@ const SubmoduleFetcher    = require("./submodule_fetcher");
 const SubmoduleConfigUtil = require("./submodule_config_util");
 const SubmoduleUtil       = require("./submodule_util");
 const TreeUtil            = require("./tree_util");
+const UserError           = require("./user_error");
 
 /**
  * Stage the specified `filename` having the specified `change` in the
@@ -141,9 +142,41 @@ function branchStatusLine(status) {
     return `On detached head ${GitUtil.shortSha(status.headCommit)}.\n`;
 }
 
-function prefixWithPound(text) {
-    return text.replace(/^/mg, "# ").replace(/^# $/mg, "#");
-}
+/**
+ * Return a string having the same value as the specified `text` except that
+ * each empty line is replaced with "#", and each non-empty line is prefixed
+ * with "# ".  The result of calling with the empty string is the empty string.
+ * The behavior is undefined unless `"" === text || text.endsWith("\n")`.
+ *
+ * @param {String} text
+ * @return {String}
+ */
+exports.prefixWithPound = function (text) {
+    assert.isString(text);
+    if ("" === text) {
+        return "";                                                    // RETURN
+    }
+    assert(text.endsWith("\n"));
+    const lines = text.split("\n");
+    const resultLines = lines.map((line, i) => {
+
+        // The split operation makes an empty line that we don't want to
+        // prefix.
+
+        if (i === lines.length - 1) {
+            return "";
+        }
+
+        // Empty lines don't get "# ", just "#".
+
+        if ("" === line) {
+            return "#";                                               // RETURN
+        }
+
+        return `# ${line}`;
+    });
+    return resultLines.join("\n");
+};
 
 /**
  * Return text describing the changes in the specified `status`.  Use the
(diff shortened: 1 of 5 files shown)


bpeabody:
Made some improvements to the syntax of the prompt.

bpeabody:
I updated https://github.com/twosigma/git-meta/issues/238 with a sample of the prompt syntax.

bpeabody:
Updated based on discussion in https://github.com/twosigma/git-meta/issues/238

node/lib/util/commit.js
```diff
@@ -1271,3 +1335,181 @@ exports.getCommitStatus = co.wrap(function *(repo, cwd, options) {
...
+        if (seen.has(tag)) {
+            throw new UserError(`${tag} was used more than once.`);
+        }
+        seen.add(tag);
+
+        const blockLines = lines.slice(start, end);
+        const message = GitUtil.stripMessageLines(blockLines);
+        if ("*" === tag) {
+            if ("" !== message) {
+                metaMessage = message;
```
novalis: do we need to check for a duplicate metaMessage?

node/lib/util/commit.js
```diff
@@ -257,6 +307,17 @@ exports.commit = co.wrap(function *(metaRepo, all, metaStatus, message) {
     const subCommits = {};
     let subsChanged = false;
     const commitSubmodule = co.wrap(function *(name) {
```
bpeabody: It's closed over enough variables: `subMessages`, `submodules`, `metaRepo`, and `subCommits`, `all`, `signature`, and `subsChanged`, that I think moving it out would not improve readability -- open to other opinions on this though!

node/lib/util/commit.js
```diff
@@ -301,23 +362,26 @@ exports.commit = co.wrap(function *(metaRepo, all, metaStatus, message) {
     const subCommitters = Object.keys(submodules).map(commitSubmodule);
```
bpeabody: The yield on arary construct is the standard way to achieve parallelism with this mechanism. I use it pretty frequently. I made the `doInParallel` method to handle situations where the normal built-in parallelism would not be able to understand and properly limit the resources we would be consuming (e.g., when forking processes).

How about a comment, e.g.:
```JavaScript
// Generate commits for submodules in parallel.
```

node/lib/util/commit.js
bpeabody: Yes, and I do here:
```JavaScript
        if (seen.has(tag)) {
            throw new UserError(`${tag} was used more than once.`);
        }
        seen.add(tag);
```

bpeabody: It gets caught the same as submodules, its tag is just "*".

novalis: oh, duh

node/lib/util/commit.js
novalis: Only subCommits and subsChanged are written to -- and the subsChanged section could be broken out into a separate function.  So I think it would be possible (and clearer) to break this out.

node/lib/util/commit.js
bpeabody: To be fair, an error message of "* was used more than once" might not be the most clear.

node/lib/util/commit.js
bpeabody: OK, will do. It will be easier to test if I break it out. Also, you are right about the `subsChanged` section; I think having it calculated like this is a legacy of an earlier version of the algorithm where it was a more complicated to compute than it is now.

node/lib/util/commit.js
novalis: There are libraries like http://caolan.github.io/async/docs.html#each which maybe make this more explicit?   I don't use enough node to really know what's idiomatic, so maybe I'm wrong to find this ugly.

node/lib/util/commit.js
bpeabody: so, I think I will change the exception to have a more intuitive message

node/lib/util/commit.js
bpeabody: I don't think anyone will fault you for finding something in JavaScript to be ugly :). The library you link is callback-based, which is a couple of generations behind what we're doing. First there was the node-style callbacks used in the `async` library you reference, which was functional but led to callback hell. Promises were next and made things much better but were still very verbose and still required lots of closures. The generators-approach lets you write async code that looks very similar to synch code (i.e., not many callbacks). Admittedly though, any "idioms" such as `yield someArray` to achieve parallelism are pretty new.

novalis: Yeah, I know the library is old -- I just googled it at random.  It just seems *weird* to have to write code like this.  But I guess I'll drop it.

bpeabody:
Apologies, @novalis: I'm in progress on the changes we've discussed so far, and pushed to this branch to "save" my work.

bpeabody:
I think I've addressed the concerns. I didn't pull `commitSubmodule` out, but I did pull out most of its complexity.

PR #253, opened 2017-04-27
Better amend

bpeabody:
- made "normal" amend able to handle more situations
- added interactive mode


node/lib/util/commit.js
```diff
@@ -1067,73 +1086,110 @@ exports.amendMetaRepo = co.wrap(function *(repo,
...
+    let result = "";
+    if (lastSignature.name() !== currentSignature.name() ||
+        lastSignature.email() !== currentSignature.email()) {
+        result += `\
+Author:    ${lastSignature.name()} <${lastSignature.email()}>
+`;
+    }
+    const time = lastSignature.when();
+    const offset = time.offset();
+    const prefix = offset < 0 ? "-" : "";
```
novalis: rename to something more-specific: "minusPrefix"?

node/lib/util/commit.js
```diff
@@ -606,182 +643,232 @@ exports.commitPaths = co.wrap(function *(repo, status, message) {
...
 
-        const relation = newSub.index.relation;
-        if (RepoStatus.Submodule.COMMIT_RELATION.SAME !== relation) {
-            newCommits[subName] = relation;
-            return;                                                   // RETURN
-        }
+    // We'll do an amend of this sub only if:
+    // - it's not gone in the index
+    // - it wasn't removed in the last commit
+    // - it wasn't added in the last commit
```
novalis: Is that really right?  What if we add it and then want to change what we've added?  Similarly what if we remove but then decide not to?

bpeabody: By "do an amend" I mean perform a commit amendment in the sub-repo; I will clarify. What you are asking should work.

node/lib/util/commit.js
```diff
@@ -1067,73 +1086,110 @@ exports.amendMetaRepo = co.wrap(function *(repo,
...
+        hour12: false,
+        year: "numeric",
+        month: "numeric",
+        day: "numeric",
+        hour: "numeric",
+        minute: "numeric",
+        second: "numeric",
+    }).format(date);
+    result += `\
+Date:      ${formatted} ${prefix}0${Math.abs(time.offset())}00
```
bpeabody: The offset appears to be in hours based on my testing.

novalis: I just (a) read the source code and tests and (b) tested manually.  It is minutes.

bpeabody: I may have been looking at the wrong thing.

bpeabody:
OK, think I got everything.

node/lib/util/commit.js
```diff
@@ -944,73 +1086,114 @@ exports.amendMetaRepo = co.wrap(function *(repo,
...
+        hour12: false,
+        year: "numeric",
+        month: "numeric",
+        day: "numeric",
+        hour: "numeric",
+        minute: "numeric",
+        second: "numeric",
+    }).format(date);
+    result += `\
+Date:      ${formatted} ${signPrefix}0${Math.abs(offset)}00
```
novalis: -1100 tho

bpeabody: whoops

bpeabody: Wait, no, that's right. I'm (lazily) rounding the offset down to the hour.

bpeabody: E.g., output from my machine in EDT:
```
# <*> enter meta-repo message above this line; delete to commit only submodules
# Date:      4/27/2017, 14:45:27 -0400
```

novalis: Your format is 0 + hour + 00.  What if the hour goes to 11?

bpeabody: OK, I am a moron. I will just kill the leading 0; will do something fancy later.

bpeabody:
I have shamefully submitted yet another fix for the TZ offset problem.

node/lib/util/commit.js
```diff
@@ -1067,73 +1086,114 @@ exports.amendMetaRepo = co.wrap(function *(repo,
...
+Author:    ${lastSignature.name()} <${lastSignature.email()}>
+`;
+    }
+    const time = lastSignature.when();
+    const signPrefix = time.offset() < 0 ? "-" : "";
+
+    //  TODO: something better than rounding offset, though I think we can live
+    //  without showing minute-specific TZ diffs for a long time.
+
+    const offset = Math.floor(time.offset() / 60);
```
novalis: FYI: This will work in this case, because we expect our offsets to be hours, but in general, Math.floor on negative numbers rounds towards negative infinity not zero.

Issue #254, opened 2017-04-27
git meta submodule status claims closed submodules are deleted


novalis:
Introduced in d42f6fb8f373.  

Demo attached.
[test.zip](https://github.com/twosigma/git-meta/files/962570/test.zip)

bpeabody:
I'll leave this issue open until I finish writing a test.

bpeabody:
done

PR #255, opened 2017-04-27
Fix `submodule status`.

bpeabody:
Addresses: https://github.com/twosigma/git-meta/issues/254


novalis:
@abliss He did mention the test as a TODO; I did a manual test with my script and it works.  I think we can merge this before having a test.

PR #263, opened 2017-05-09
Cache and uncache submodule states as needed.

bpeabody:
Also upgraded nodegit.  Now we can cache this state explicitly just when we
need to.  Note that using the cache all the time causes operations like
`rebase` to fail.

Updated `read_repo_ast_util` test driver because libgit2 now creates a
different merge commit error message.

### DIFF
--- a/node/lib/util/rebase_util.js
+++ b/node/lib/util/rebase_util.js
@@ -684,11 +684,13 @@ up-to-date.`);
                 yield NodeGit.AnnotatedCommit.fromRef(metaRepo, currentBranch);
         const ontoAnnotatedCommit =
                       yield NodeGit.AnnotatedCommit.lookup(metaRepo, commitId);
-        const rebase = yield NodeGit.Rebase.init(metaRepo,
-                                                 fromAnnotedCommit,
-                                                 ontoAnnotatedCommit,
-                                                 null,
-                                                 null);
+        const rebase = yield SubmoduleUtil.cacheSubmodules(metaRepo, () => {
+            return NodeGit.Rebase.init(metaRepo,
+                                       fromAnnotedCommit,
+                                       ontoAnnotatedCommit,
+                                       null,
+                                       null);
+        });
         console.log(`Rewinding to ${colors.green(commitId.tostrS())}.`);
         yield callNext(rebase);
         return rebase;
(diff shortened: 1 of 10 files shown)


node/lib/util/rebase_util.js
```diff
@@ -707,6 +709,7 @@ up-to-date.`);
 exports.abort = co.wrap(function *(repo) {
     assert.instanceOf(repo, NodeGit.Repository);
 
+    repo.submoduleCacheAll();
     const rebase = yield NodeGit.Rebase.open(repo);
     rebase.abort();
```
novalis: should we clear after aborting the rebase?

node/lib/util/status_util.js
```diff
@@ -387,11 +388,13 @@ exports.getRepoStatus = co.wrap(function *(repo, options) {
             const treeId = head.treeId();
             tree = yield NodeGit.Tree.lookup(repo, treeId);
         }
+        repo.submoduleCacheAll();
         const status = yield DiffUtil.getRepoStatus(repo,
                                                     tree,
                                                     options.paths,
                                                     options.ignoreIndex,
                                                     options.showAllUntracked);
+        repo.submoduleCacheClear();
```
novalis: Do we really need to clear here? We haven't changed the state of the repo.

bpeabody: Clearing the cache also turns it off. I think it's a good idea to leave the cache "off" at the end of this.

node/lib/util/rebase_util.js
bpeabody: Yes, but might as well do it right after `open`.

node/lib/util/status_util.js
novalis: I guess the idea is that the cache is only useful during certain single operations?  If so, can we in general have wrappers for those operations which cache/uncache around them?

bpeabody: Yes, I wasn't sure it would be the case (that we would always cache during a single operation), but you're right.

bpeabody: OK, there is at least one case (checkout) where we'd need to cache/uncache more than once. I'll refactor the other cases.

bpeabody:
OK, wrapped up all the one-op operations.

PR #271, opened 2017-05-22
Stash

bpeabody:


node/lib/util/tree_util.js
```diff
@@ -191,3 +193,83 @@ exports.writeTree = co.wrap(function *(repo, baseTree, changes) {
...
+};
+
+/**
+ * Return a map from path to `Change` for the working directory of the
+ * specified `repo` having the specified `status`.  If the specified `all` is
+ * true, include unstaged changes.
+ *
+ * @param {NodeGit.Repository} repo
+ * @param {RepoStatus}         status
+ * @param {Boolean}            all
```
bpeabody: yeah, you're right. I saw an `all` option in `git stash`, but it means to also include ignored files

node/lib/cmd/stash.js
```diff
@@ -0,0 +1,122 @@
...
+exports.description =`
+Provide commands for saving and restoring the state of the monorepo.
+Note that 'stash' affects only the open repositories of *sub-repos*; as
+currenty implemented, the meta-repo itself is not affected, including staged
+and unstaged commits to the sub-repos.`;
+
+exports.configureParser = function (parser) {
+
+    parser.addArgument("type", {
+        help: "'save' to save a stash, 'pop' to restore; 'save' is default",
```
bpeabody: I think you're wrong in this one though:

```
save [-p|--patch] [-k|--[no-]keep-index] [-u|--include-untracked] [-a|--all] [-q|--quiet] [<message>]
           Save your local modifications to a new stash, and run git reset --hard to revert them. The <message> part is optional and gives the description along
           with the stashed state. For quickly making a snapshot, you can omit both "save" and <message>, but giving only <message> does not trigger this action
           to prevent a misspelled subcommand from making an unwanted stash.
```
vs.
```
create
           Create a stash (which is a regular commit object) and return its object name, without storing it anywhere in the ref namespace. This is intended to be
           useful for scripts. It is probably not the command you want to use; see "save" above.
```

node/lib/util/stash_util.js
```diff
@@ -0,0 +1,391 @@
...
+        // Record the values we've created.
+
+        subChanges[name] = new TreeUtil.Change(
+                                            stashId,
+                                            NodeGit.TreeEntry.FILEMODE.COMMIT);
+    }));
+    const head = yield repo.getHeadCommit();
+    const headTree = yield head.getTree();
+    const subsTree = yield TreeUtil.writeTree(repo, headTree, subChanges);
+    const stashId = yield NodeGit.Commit.create(repo,
```
bpeabody: It's normal to me; I find it a little bothersome to write but pleasant to read.

No desire to extend column length at the moment.

node/lib/cmd/stash.js
novalis: Oops, yeah.

bpeabody:
I think I got everything.

Issue #270, opened 2017-05-22
Add command for changing directory to the root of the git meta clone


william-clarke:
`git meta root` or similar

william-clarke:
This bash function works, a little hacky though:
`gmcd () {
  cdpath=$(git rev-parse --show-cdup)
  cdpath+="../"
  cdpath+=$(git -C $cdpath rev-parse --show-cdup)
  cd $cdpath
}`

william-clarke:
Actually, that only works from the codebase. This works for both: 
`gmroot () {
  cdpath=$(git rev-parse --show-cdup)
  if [ ! -e "${cdpath}.gitmodules" ] ; then
    cdpath+="../"
    cdpath+=$(git -C $cdpath rev-parse --show-cdup)
  fi
  cd "$cdpath"
}
`

PR #272, opened 2017-05-22
Added the `root` command.

bpeabody:
Prints the root of the workdir of a meta-repo.

Addresses: https://github.com/twosigma/git-meta/issues/270


node/lib/cmd/root.js
```diff
@@ -0,0 +1,62 @@
...
+};
+
+/**
+ * Execute the `add` command according to the specified `args`.
+ *
+ * @async
+ * @param {Object}   args
+ * @param {String[]} args.paths
+ */
+exports.executeableSubcommand = function () {
```
jhedwardyang: For similar style with other commands, can we have this return a generator wrapped in a Promise by co.wrap?

bpeabody: I'd have to do something a little hacky to stop JSHint from complaining about a lack of `yield` in the body of the generator.

jhedwardyang: I'm okay with this merged in the current state then.

jhedwardyang:
Can you update this review to get Travis running on it? Thanks!

PR #278, opened 2017-05-30
Fix flaky continue tests.

bpeabody:
Because there was no branch pointing to the "onto" commit configured in the
some submodule tests, it was sometimes being GC'd by the process I use to write
out repos (which explicitly calls GC at the end).


bpeabody:
Looks like this bombed due to a GitHub/TravisCI problem; will re-push.

bpeabody:
Not sure what was going on with Travis there, but this is finally passed.

PR #301, opened 2017-07-13
make thread safe

novalis:


node/lib/git-meta.js
```diff
@@ -58,6 +58,11 @@ const status     = require("./cmd/status");
 const UserError  = require("./util/user_error");
 const version    = require("./cmd/version");
 
+// see https://github.com/nodegit/nodegit/issues/827 -- this is required
+// to prevent random hard crashes with e.g. parallelism in index operations.
+// Eventually, this will be nodegit's default.
+NodeGit.setThreadSafetyStatus(NodeGit.THREAD_SAFETY.ENABLED_FOR_ASYNC_ONLY);
```
bpeabody: I don't think `NodeGit` has been required (indeed, it fails a the linter).

bpeabody:
@novalis : were you able to verify that setting this flag fixed the problem? I don't think the code in its current form works.

node/lib/git-meta.js
novalis: Was rushing out the door, sorry.  Should have tested.

PR #323, opened 2017-07-28
Support alternate object directories in hook

novalis:
Git 2.11 quarantines incoming packs.  This means that they're not in
the normal object database -- they're in an alternate object database
untli the push is accepted.  Git tells us about the alternate ODB
through environment variables.  We need to use this alternate ODB
to look up objects in our hook.


novalis:
Some local testing done.

PR #328, opened 2017-08-02
Force checkout

bpeabody:
- adds `force` option to checkout
- refuse to overwrite submodules with local commits unless forced

### DIFF
--- a/node/lib/util/checkout.js
+++ b/node/lib/util/checkout.js
@@ -339,13 +339,19 @@ A branch named ${colors.red(name)} already exists.`);
             const commit = yield repo.getCommit(annotated.id());
             result.commit = commit;
 
-            // Check to see if the commit refers to a branch name.
+            // Check to see if the commit refers to a branch name.  You would
+            // think that 'findBranch' would return only branches, but it also
+            // returns 'FETCH_HEAD', which is *not a branch*.  "HEAD",
+            // unfortunately, claims to be a branch, and must be special-cased.
 
-            const branch = yield GitUtil.findBranch(repo, committish);
-            if (null !== branch) {
-                committishBranch = branch;
-                if(!branch.isRemote()) {
-                    result.switchBranch = committish;
+            if ("HEAD" !== committish) {
+                const branch = yield GitUtil.findBranch(repo, committish);
+                if (null !== branch &&
+                    (branch.isBranch() || branch.isRemote())) {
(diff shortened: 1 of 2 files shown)


node/lib/util/checkout.js
```diff
@@ -116,49 +118,62 @@ exports.checkoutCommit = co.wrap(function *(metaRepo, commit) {
...
-    if (null !== metaError) {
-        errors.push(`Unable to check out meta-repo: ${metaError}.`);
-    }
-
-    // Try the submodules; store the opened repos and loaded commits for use
-    // in the actual checkout later.
-
     const cache = {};  // name to { repo, commit}
 
+    // Load repo/commit cache
```
novalis: This method is getting long -- when you have named sections, it's often better to split them out into named methods.  Would that be possible here?

bpeabody: Yes, that's fair.

node/lib/util/checkout.js
```diff
@@ -116,49 +118,62 @@ exports.checkoutCommit = co.wrap(function *(metaRepo, commit) {
...
-    if (0 !== errors.length) {
-        throw new UserError(errors.join("\n"));
+        const metaError = yield dryRun(metaRepo, commit);
+        if (null !== metaError) {
+            errors.push(`Unable to check out meta-repo: ${metaError}.`);
+        }
+
+        // Try the submodules; store the opened repos and loaded commits for
+        // use in the actual checkout later.
+
```
novalis: nit: too many blank lines

node/lib/util/checkout.js
```diff
@@ -116,49 +118,62 @@ exports.checkoutCommit = co.wrap(function *(metaRepo, commit) {
...
+
+        yield open.map(co.wrap(function *(name) {
+            // Open repo but not alive on this commit.
```
novalis: What does "alive" mean?

node/lib/util/checkout.js
```diff
@@ -414,6 +429,7 @@ Cannot setup tracking information; starting point is not a branch.`);
  * - configure the new branch to have the specified `newBranch.tracking`
  *   tracking branch
  * - make the specified `switchBranch` the current branch
+ * - overwrite local changes unless `true === force`
```
novalis: this comment seems funny -- do you mean "if" instead of "unless"?

PR #332, opened 2017-08-03
Fix crash when no tracking info on pull

bpeabody:
Addresses: https://github.com/twosigma/git-meta/issues/285

### DIFF
--- a/node/lib/cmd/pull.js
+++ b/node/lib/cmd/pull.js
@@ -97,8 +97,10 @@ exports.executeableSubcommand = co.wrap(function *(args) {
         throw new UserError("Non-rebase pull not supported.");
     }
 
+    // TODO: move the following code into `util/push.js` and add test
+
     const branch = yield repo.getCurrentBranch();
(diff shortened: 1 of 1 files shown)


node/lib/cmd/pull.js
```diff
@@ -97,8 +97,10 @@ exports.executeableSubcommand = co.wrap(function *(args) {
         throw new UserError("Non-rebase pull not supported.");
     }
 
+    // TODO: move the following code into `util/push.js` and add test
+
     const branch = yield repo.getCurrentBranch();
-    const tracking = yield GitUtil.getTrackingInfo(branch);
+    const tracking = (yield GitUtil.getTrackingInfo(branch)) || {};
```
novalis: Does this mean that getTrackingInfo returns the wrong thing?  That is, should it return {} instead of null?

bpeabody: No, it's supposed to return `null` when there is no tracking info.

Issue #321, opened 2017-07-26
`rebase` appears to open submodules unnecessarily


bpeabody:
e.g.:
```
Submodule xyz: opening
Submodule xyz: starting rebase; rewinding to b9ab66e449cea6ba7296d968a9bbc60ebec87c03.
Submodule xyz: finished rebase.
```

PR #333, opened 2017-08-03
Simplified rebase

bpeabody:


bpeabody:
I was not able to recreate https://github.com/twosigma/git-meta/issues/324, but the codepath through with that NPE could happen no longer exists, so I believe this PR will fix that problem.

bpeabody:
Should also address https://github.com/twosigma/git-meta/issues/321

node/lib/util/rebase_util.js
```diff
@@ -120,193 +108,106 @@ const callFinish = co.wrap(function *(repo, rebase) {
...
- * @param {SubmoduleFetcher}   fetcher
+ * @param {NodeGit.Index} index
+ * @return {Boolean}
  */
-function SubmoduleRebaser(submoduleName, repo, submoduleCommits, getFetcher) {
-    assert.isString(submoduleName);
-    assert.instanceOf(repo, NodeGit.Repository);
-    assert.isObject(submoduleCommits);
-    assert.isFunction(getFetcher);
+const stageAll = co.wrap(function *(index) {
```
novalis: Isn't this Index#updateAll?

novalis:
Looks like there are also some test failures.

bpeabody:
Two test failures, one I believe from travis flakiness and one from what appears to be flakiness of one of my tests (I've seen it fail there before). Created https://github.com/twosigma/git-meta/issues/335 to address that.

node/lib/util/rebase_util.js
bpeabody: You might think so, but `updateAll` doesn't update "all", it updates all paths that match an array of wildcards.

bpeabody:
Addressed feedback and squashed the three simplifications into one commit.

node/lib/util/rebase_util.js
novalis: But "*" would work, right?

bpeabody:
Ping on this one. I think I've addressed everything; it should address most of our current rebase issues.

PR #334, opened 2017-08-03
Fix alternate ODB code

novalis:
Oops.  This code was in the wrong place.


bpeabody:
Also, you're jhint check is failing.

novalis:
Apparently messing with GIT_OBJECT_DIRECTORY is bad (even when I put it back after the test).  Doing GIT_ALTERNATE_OBJECT_DIRECTORIES instead.

novalis:
Remaining test failure is known flakiness.

novalis:
ping on this; it's blocking the hook from being used with a recent gitlab.

PR #338, opened 2017-08-04
respect user's pull.rebase and branch.<name>.rebase settings

### LINKED ISSUE #307: git-meta ignores pull.rebase setting
$ git config --get pull.rebase
true
$ git meta pull upstream master
"Non-rebase pull not supported"

novalis:
fixes #307.

### DIFF
--- a/node/lib/cmd/pull.js
+++ b/node/lib/cmd/pull.js
@@ -82,7 +82,7 @@ used if not specified`,
  *
  * @async
  * @param {Object}  args
- * @param {Boolean} args.any
+ * @param {Boolean} args.rebase
  * @param {String}  repository
  * @param {String}  [source]
  */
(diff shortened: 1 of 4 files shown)


node/lib/cmd/pull.js
```diff
@@ -78,11 +78,47 @@ used if not specified`,
...
+ * via config).
+ *
+ * @param {Object} args
+ * @param {Boolean} args.rebase
+ * @param {Nodegit.Repository} repo
+ * @param {NodeGit.Branch} branch
+ * @async
+ * @return bool
+ */
+exports.userWantsRebase = co.wrap(function*(args, repo, branch) {
```
bpeabody: Can you move this to 'lib/until/pull.js'?  I generally try to keep the code in cmd to a bare minimum and export nothing but the handlers, partially to minimize startup time and also to avoid library code from becoming mingled with argument handlers.

node/lib/cmd/pull.js
```diff
@@ -78,11 +78,47 @@ used if not specified`,
...
+        const configured = yield config.getStringBuf(branchConfigVar);
+        return (configured === "true" || configured === "yes" ||
+                configured === "on");
+    } catch (e) {
+        // no branch config, try global config
+    }
+
+    try {
+        const configured = yield config.getStringBuf("pull.rebase");
+        return (configured === "true" || configured === "yes" ||
```
bpeabody: Could you factor out the test against the three strings? It's unfortunate that libgit2 doesn't do this.

PR #381, opened 2017-10-03
Optimized getSubmoduleChanges

bpeabody:
Now computes via `NodeGit.Diff`, which should be much faster than doing
it "by hand".

### DIFF
--- a/node/lib/util/submodule_util.js
+++ b/node/lib/util/submodule_util.js
@@ -305,48 +305,81 @@ exports.getSubmoduleRepos = co.wrap(function *(repo) {
 });
 
 /**
- * Return a summary of the submodules changed by the specified `commitId` in
- * the specified `repo`.
+ * Return a summary of the submodule SHAs changed by the specified `commitId` in
+ * the specified `repo`, and flag denoting whether or not the `.gitmodules`
+ * file was changed.
  *
  * @asycn
  * @param {NodeGit.Repository} repo
  * @param {NodeGit.Commit}     commit
  * @return {Object}
- * @return {Set(String)}  return.added    paths to added submodules
- * @return {Set(String)}  return.changed  paths to changed submodule
- * @return {Set(String)}  return.removed  paths to removed submodules
+ * @return {Object}  return.added    map from path to SHA
+ * @return {Object}  return.changed  map from path to new and old SHAs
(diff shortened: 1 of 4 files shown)


node/lib/util/submodule_util.js
```diff
@@ -305,48 +305,80 @@ exports.getSubmoduleRepos = co.wrap(function *(repo) {
...
+            case DELTA.DELETED: {
+                const oldFile = delta.oldFile();
+                const path = oldFile.path();
+                if (COMMIT === oldFile.mode()) {
+                    result.removed[oldFile.path()] = oldFile.id().tostrS();
+                }
+                else if (SubmoduleConfigUtil.modulesFileName === path) {
+                    result.modulesFileChanged = true;
+                }
+            } break;
```
novalis: case DELTA.RENAMED.  case DELTA.COPIED.

node/lib/util/submodule_util.js
```diff
@@ -305,48 +305,80 @@ exports.getSubmoduleRepos = co.wrap(function *(repo) {
...
  * @asycn
  * @param {NodeGit.Repository} repo
  * @param {NodeGit.Commit}     commit
  * @return {Object}
- * @return {Set(String)}  return.added    paths to added submodules
- * @return {Set(String)}  return.changed  paths to changed submodule
- * @return {Set(String)}  return.removed  paths to removed submodules
+ * @return {Object}  return.added    map from path to SHA
+ * @return {Object}  return.changed  map from path to new and old SHAs
+ * @return {Object}} return.removed  map from epath to SHA
```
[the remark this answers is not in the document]
bpeabody: I don't need the SHA on removed right now, but it's readily-available so I thought I'd go ahead and include it; I think I'll need it if I want to use this API to replace some more-expensive ones I'm using in, e.g., `git meta status`.

node/lib/util/submodule_util.js
bpeabody: I put in a TODO for now. Not sure if these crop up with submodules or how I'd handle yet; I need to do more testing.

novalis: IIRC, I have seen them while doing some manual gitlab testing at one point.  But I could be misremembering.  At the very least, we ought to die when we see them so that we don't just skip them.

PR #387, opened 2017-10-07
Add support for actual result mapping in tests.

bpeabody:
This allows testers to massage the `RepoAST` objects read from disk
before they are compared to the expected result.  Useful when, e.g., ref
name, may contain physical commit shas that need to be mapped to the
logical shas referenced in the expected result.


node/lib/util/repo_ast_test_util.js
```diff
@@ -230,6 +239,12 @@ exports.testMultiRepoManipulator =
     else {
         assert.isFunction(options.expectedTransformer);
     }
+    if (!("actualTransformer" in options)) {
```
novalis: nit: swap if/else clauses and make if statement positive rather than negative.

bpeabody: I entered an issue to remind myself to fix this: https://github.com/twosigma/git-meta/issues/389

PR #408, opened 2017-11-07
gitmeta add-submodules should sort .gitmodels

s-xiang:


novalis:
This doesn't actually solve the problem, because the sorting doesn't take into account the underscores.

bpeabody:
That will be done in a local patch to the sorting logic. Hooking up the add operation to sorting is something that makes sense generally.

PR #411, opened 2017-11-09
Fix sorting of in add-submodule

bpeabody:
We were sorting and writing after we updated the index.

### DIFF
--- a/node/lib/util/add_submodule.js
+++ b/node/lib/util/add_submodule.js
@@ -65,19 +65,14 @@ exports.addSubmodule = co.wrap(function *(repo, url, filename, importArg) {
     }
     const modulesPath = path.join(repo.workdir(),
                                   SubmoduleConfigUtil.modulesFileName);
-    fs.appendFileSync(modulesPath, `\
-[submodule "${filename}"]
-    path = ${filename}
-    url = ${url}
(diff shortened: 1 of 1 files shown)


node/lib/util/add_submodule.js
```diff
@@ -65,20 +65,15 @@ exports.addSubmodule = co.wrap(function *(repo, url, filename, importArg) {
...
-    url = ${url}
-`);
     const index = yield repo.index();
-    yield index.addByPath(SubmoduleConfigUtil.modulesFileName);
-    yield index.write();
-
-    // sort all the urls since new modules get added
     const urls = yield SubmoduleConfigUtil.getSubmodulesFromIndex(repo, index);
+    urls[filename] = url;
     const newConf = SubmoduleConfigUtil.writeConfigText(urls);
```
jhedwardyang: I ran this and it seems to remove the newline at the end of .gitmodules.

jhedwardyang:
arggg. I was 25 seconds too late

node/lib/util/add_submodule.js
bpeabody: There should be a local patch to perform our site-local sorting logic.

PR #426, opened 2017-11-27
commit-shadow

bpeabody:
Created command for making a "shadow" commit -- recording local changes without affecting the index or HEAD. Also fixed some problems that would have kept it from working (and might have caused problems for stash as well).


node/lib/cmd/commit-shadow.js
```diff
@@ -0,0 +1,92 @@
...
+exports.helpText =
+    `Create a "shadow" commit, leaving index and HEAD unchanged.`;
+
+/**
+ * description of the `commit-shadow` comand
+ *
+ * @property {String}
+ */
+exports.description  = `Create a "shadow" commit containing all local
+modifications (including untracked files) to all sub-repos and print its SHA.
```
novalis: Not sure we ought to include untracked files -- we've definitely seen tests fail because of e.g. a foo.orig file, and this seems like a good way to get annoying-to-track-down bug reports.

node/lib/cmd/commit-shadow.js
```diff
@@ -0,0 +1,92 @@
...
+If there are no local modifications, print the SHA of HEAD.  Do not modify the
+index or update HEAD to point to the created commit.  Note that this command
+ignores changes to the meta-repo.  Note that this command is meant for
```
novalis: Wait, I thought it handled added/removed submodules?

node/lib/cmd/commit-shadow.js
bpeabody: I will make it an option, but I think in our intended use-case -- replicating the user's local worktree for, e.g., remote build -- we'll want them.

node/lib/cmd/commit-shadow.js
bpeabody: Sorry, I should clarify that it ignore non-submodule changes to the meta-repo. The reasons for this are (a) it's faster if we don't have to do the equivalent of a `git-status` in the meta-repo and (b) I have some work to do in my own `getRepoStatus` method for this case (where `ignoreIndex` is true) so that it's doesn't see files in new submodules as untracked.

abliss:
I'm pretty sure "vats build" ignores untracked files, so users might find
this confusing--especially if they never use local make and are accustomed
to leaving crud in their dir. But I agree that including untracked files
probably makes more sense so maybe they can adjust.

What about skipping "large files" (for some threshold) by default though?
people may have large garbage hanging around (e.g. core dumps) and it'd be
a shame to add these to the object db and push them.

bpeabody:
I would hope that transient (large) files are ignored.

For now, I'll add an `--include-untracked` option (default, false). If we decide to start using it and have an issue with large files, I'll add a "max size for untracked" option.

abliss:
I like your solution. I suspect a lot of people occasionally leave
transient large untracked files around without gitignoring them.

bpeabody:
Updated with requested changes including the `--include-untracked` option (default: false).

PR #445, opened 2017-12-11
Added garbage collector for synthetic refs.

zokir:
First draft. Let me know what you think @novalis. I will write some function level docs for now as well. 
Forgive my javascript :) 

Some details:
**Important branches:** Are persistent branches that assumed to be around for a long time, like master or team branches. What that gives us is that we can remove synthetic refs for their parents since descendant assumed to be around. The problem we are having right now is how to come up with that list.

After some discussion with David seems like we might not need the concept of important branches. Already wrote the code tho, so might be useful in the future. 

How I tested it:
Setup a meta repo with two submodules. 
Had two important branches (master etc.), and two unimportant branches(user branches). 

1) Created couple of commits in important branches, ran synthetic-gc, observed that only the lowest descendants had synthetic refs. Checked the case where lowest descendant is in another branch.

2) Created couple of commits in unimportant and important branches, ran synthetic-gc with a specific date that is later than those commits. Observed that synthetic refs for unimportant commits that are older than specified threshold date got deleted.

### DIFF
--- a/node/lib/synthetic_gc.js
+++ b/node/lib/synthetic_gc.js
@@ -1,0 +1,231 @@
+#! /usr/bin/env node
+
+'use strict';
+
+const assert  = require("chai").assert;
+const co = require("co");
+const NodeGit = require("nodegit");
+const SyntheticBranchUtil = require("./util/synthetic_branch_util");
+const GitUtil = require("./util/git_util");
+const SubmoduleUtil = require("./util/submodule_util");
+const ArgumentParser = require('argparse').ArgumentParser;
+
+const SYNTHETIC_BRANCH_BASE = "refs/commits/";
+
+let SIMULATION = true; // dont' actually delete refs, just output
+
+let visited = {}; // visited commmits.
+
+function parseOptions() {
+    let parser = new ArgumentParser({
+        version: '0.1',
+        addHelp:true,
+        description: "Old synthetic refs removal. By default removes all synthetic refs that are older than 6 moths." 
+    });
+
+    parser.addArgument(
+        [ '-d', '--date' ],
+        { help: 'Date to use as a threshold for removing old synthetic refs.[6 month default]' }
+    );
+
+    parser.addArgument(
+        [ '-f', '--force' ],
+        {
+            help: 'Actually remove synthetic refs. By Default we only list what refs we are about to remove.',
+            action: 'storeTrue',
+            defaultValue: false,
+        }
+    );
+    
+    return parser.parseArgs();
+}
+
+
+function* removeSyntheticRef(repo, commit) {
+    const refPath = SyntheticBranchUtil.getSyntheticBranchForCommit(commit);
+
+    console.log("Removing ref: " + refPath);
+
+    if (SIMULATION) {
+        return;
+    }
+    
+    return yield GitUtil.removeRemoteRef(repo, refPath);
+}
+
+// Go through the parents of 'commit' and remove synthetic reference recusively if they satisfy 'isDeletable'. 
+function* recursiveSyntheticRefRemoval(repo, commit, isDeletable, existingReferences) { 
+
+    assert.instanceOf(repo, NodeGit.Repository);
+    assert.instanceOf(commit, NodeGit.Commit);
+
+    if (commit.parentcount() === 0) {
+        return;
+    }
+
+    const parents = yield commit.getParents(commit.parentcount());
+    yield parents.map(function *(parent) {
+        if (parent in visited) {
+            return;
+        }
+        visited[parent] = 1;
+
+        if (isDeletable(parent) && existingReferences.includes(parent.sha())) {
+            yield removeSyntheticRef(repo, parent);
+        }
+        return yield *recursiveSyntheticRefRemoval(repo, parent, isDeletable, existingReferences);
+    });
+
+    return;
+}
+
+    // Clean up all redundant 
+function* cleanUpRedundant(repo, roots, predicate) {
+
+   for (let subName in roots) {
+       const subRepo = yield SubmoduleUtil.getRepo(repo, subName);
+
+       let existingReferences = yield getSyntheticRefs(subRepo);
+
+       for (let subCommit of roots[subName]) {
+           yield recursiveSyntheticRefRemoval(subRepo, subCommit, predicate, existingReferences); 
+       }
+   }
+}
+
+function* getSyntheticRefs(subRepo) {
+
+    const syntheticRefExtractor = function(value) {
+          if (value && value.includes(SYNTHETIC_BRANCH_BASE) ) {
+              return value.split("\t")[0];
+          }
+          return null;
+    }
+
+    let references = yield GitUtil.getRefs(subRepo);
+
+    return references.map(syntheticRefExtractor).filter(commit => commit);
+}
+
+    // Clean up all refs that satisfy 'isOldCommit', but not part of 'roots'.
+function* cleanUpOldRefs(repo, roots, isOldCommit) {
+
+   for (let subName in roots) {
+       const subRepo = yield SubmoduleUtil.getRepo(repo, subName);
+      
+       const reservedCommits = roots[subName];
+
+       let rawShaSubRoots = Array.from(reservedCommits).map(function(commit) {
+                            return commit.sha();
+                        });
+
+       let references = yield getSyntheticRefs(subRepo);
+
+       // filter out all the references from rootA
+       references = references.filter(ref => !rawShaSubRoots.includes(ref)); 
+
+
+       // filter out all the references younger than 6 months
+       let refsToDelete = [];
+       for (let ref in references) {
+           // could not figure out how to call generators withing filter properly, so cant be fancy
+           const actualCommit = yield subRepo.getCommit(references[ref]);
+           if (isOldCommit(actualCommit)) {
+               refsToDelete.push(actualCommit);
+           }
+       }
+
+        
+       yield* refsToDelete.map(ref => removeSyntheticRef(subRepo, ref));
+   }
+}
+
+let lessThanDate = function(thresHold) {
+    return function(input) {
+        return input.date() < thresHold;
+    }
+}
+
+function getThresholdDate(args) {
+
+    let date = new Date();
+
+    if (args.date != undefined) {
+        date = new Date(args.date);
+    } else {
+        const THRESHOLD_MONTHS = 6;
+        date.setMonth(date.getMonth() - THRESHOLD_MONTHS); 
+    }
+
+    console.log("Using: " + date + ", as threshold date.");
+
+    return date;
+}
+
+// Fetch all important refs from out repo.
+function* populateRoots(repo) {
+
+    // For now, we are using heads/master as an important root.  
+    // 'important root' - means the root that most likey be around, so that we can remove all parent synthetic refs.
+    // This is not really necessary, more like optimization, 
+    // unless there is an important ref out there that was not updated for so long.
+    const IMPORTANT_REFS = ['refs/heads/fancy_branch', 'refs/heads/master'];
+
+    let classAroots = {}; // roots that we can rely on to be around, master or team branches
+    let classBroots = {}; // root that can go anyway, like users branches
+
+    const submodules = yield SubmoduleUtil.getSubmoduleNames(repo);
+
+    const refs = yield repo.getReferenceNames(NodeGit.Reference.TYPE.LISTALL);
+    for (let ref in refs) {
+        ref = refs[ref]
+        const refHeadCommit = yield repo.getReferenceCommit(ref);
+
+        const tree = yield refHeadCommit.getTree();
+
+        // This could have been fancy if we were running gc per meta change
+        //const submodules = yield SubmoduleUtil.getSubmoduleChanges(repo,
+        //                                                           refHeadCommit);
+
+        yield submodules.map(function*(subName) {
+            const subRepo = yield SubmoduleUtil.getRepo(repo, subName);
+            const subSha = yield tree.entryByPath(subName);
+            const subCommit = yield subRepo.getCommit(subSha.sha());
+
+            if (IMPORTANT_REFS.includes(ref)) {
+                if (!(subName in classAroots)) {
+                    classAroots[subName] = new Set();
+                }
+                classAroots[subName].add(subCommit);
+            } else {
+                if (!(subName in classBroots)) {
+                    classBroots[subName] = new Set();
+                }
+                classBroots[subName].add(subCommit);
+            }
+        })
+    }
+
+    return classAroots;
+}
+
+let runIt = co.wrap(function *(args) {
+
+    const repo = yield GitUtil.getCurrentRepo();
+    const classAroots = yield populateRoots(repo);
+
+    console.log("Looking for removal of redundant synthetic refs. (parent refs of persistent branches).")
+    yield cleanUpRedundant(repo, classAroots, function(commit) { return true; }); 
+
+    console.log("Looking for removal of old synthetic refs.")
+    const isOldCommit = lessThanDate(getThresholdDate(args));
+    yield cleanUpOldRefs(repo, classAroots, isOldCommit); 
+});
+
+const args = parseOptions();
+
+if (args.force) {
+    SIMULATION = false;
+}
+
+runIt(args);
--- a/node/lib/synthetic_gc.js
+++ b/node/lib/synthetic_gc.js
@@ -16,6 +16,11 @@ let SIMULATION = true; // dont' actually delete refs, just output
 
 let visited = {}; // visited commmits.
 
+/**
+ * Parse command line options.
+ *
+ * @return {Object}
+ */
 function parseOptions() {
     let parser = new ArgumentParser({
         version: "0.1",
@@ -43,7 +48,13 @@ function parseOptions() {
     return parser.parseArgs();
 }
 
-
+/**
+ * Remove synthetic ref corresponding to specified `commit` in the specified
+ * `repo`.
+ *
+ * @param {NodeGit.Repo}   repo
+ * @param {NodeGit.Commit} commit
+ */
 function* removeSyntheticRef(repo, commit) {
     const refPath = SyntheticBranchUtil.getSyntheticBranchForCommit(commit);
 
@@ -55,8 +66,16 @@ function* removeSyntheticRef(repo, commit) {
     return yield GitUtil.removeRemoteRef(repo, refPath);
 }
 
-// Go through the parents of `commit` and remove synthetic reference recusively 
-// if they satisfy `isDeletable`. 
+/**
+ * Go through the parents of `commit` of the specified `repo` and remove
+ * synthetic reference recusively if they satisfy `isDeletable` and not part of
+ * `existingReferences`. 
+ *
+ * @param {NodeGit.Repo}   repo
+ * @param {NodeGit.Commit} commit
+ * @param {Function}       isDeletable 
+ * @param {String[]}       existingReferences
+ */
 function* recursiveSyntheticRefRemoval(repo, commit, isDeletable, 
                                        existingReferences) { 
 
@@ -84,6 +103,12 @@ function* recursiveSyntheticRefRemoval(repo, commit, isDeletable,
     return;
 }
 
+/**
+ * Return all available synthetic refs within specifed `subRepo`.
+ *
+ * @param {NodeGit.Repo}   subRepo
+ * @return {String[]}
+ */
 function* getSyntheticRefs(subRepo) {
 
     const syntheticRefExtractor = function(value) {
@@ -98,7 +123,18 @@ function* getSyntheticRefs(subRepo) {
     return references.map(syntheticRefExtractor).filter(commit => commit);
 }
 
-    // Clean up all redundant 
+/**
+ * Delete all redundant synthetic refs within specified 'repo' satisfying
+ * `predicate` by recursively iterating over parents of the specified `roots`. 
+ *
+ * Synthetic ref is considered to be redundant if its commit is reachable from
+ * descendant who is guaranteed to be around - i.e part of a persistent roots
+ * ('roots' here).
+ *
+ * @param {NodeGit.Repo}   repo
+ * @param {Object[]}       roots
+ * @param {Function}       predicate
+ */
 function* cleanUpRedundant(repo, roots, predicate) {
 
    for (let subName in roots) {
@@ -113,7 +149,14 @@ function* cleanUpRedundant(repo, roots, predicate) {
    }
 }
 
-    // Clean up all refs that satisfy `isOldCommit`, but not part of `roots`.
+/**
+ * Delete all synthetic refs within specified `repo` that satisfy `isOldCommit`,
+ * that not part of specified `roots`.
+ *
+ * @param {NodeGit.Repo}   repo
+ * @param {Object[]}       roots
+ * @param {Function}       isOldCommit 
+ */
 function* cleanUpOldRefs(repo, roots, isOldCommit) {
 
    for (let subName in roots) {
@@ -164,7 +207,16 @@ function getThresholdDate(args) {
     return date;
 }
 
-// Fetch all important refs from out repo.
+/**
+ * Fetch all refs that are considered to be persistent within the specified
+ * `repo`. 
+ *
+ * Return value is a mapping between submodule name and collection of persistent
+ * refs within that submodules.
+ *
+ * @param {NodeGit.Repo}   repo
+ * @return {Object[]}      
+ */
 function* populateRoots(repo) {
 
     // For now, we are using heads/master as an important root.  
--- a/node/lib/synthetic_gc.js
+++ b/node/lib/synthetic_gc.js
@@ -1,6 +1,6 @@
 #! /usr/bin/env node
 
-'use strict';
+"use strict";
 
 const assert  = require("chai").assert;
 const co = require("co");
@@ -8,7 +8,7 @@ const NodeGit = require("nodegit");
 const SyntheticBranchUtil = require("./util/synthetic_branch_util");
 const GitUtil = require("./util/git_util");
 const SubmoduleUtil = require("./util/submodule_util");
-const ArgumentParser = require('argparse').ArgumentParser;
+const ArgumentParser = require("argparse").ArgumentParser;
 
 const SYNTHETIC_BRANCH_BASE = "refs/commits/";
 
@@ -18,21 +18,24 @@ let visited = {}; // visited commmits.
 
 function parseOptions() {
     let parser = new ArgumentParser({
-        version: '0.1',
+        version: "0.1",
         addHelp:true,
-        description: "Old synthetic refs removal. By default removes all synthetic refs that are older than 6 moths." 
+        description: `Old synthetic refs removal. By default removes all 
+           synthetic refs that are older than 6 moths.` 
     });
 
     parser.addArgument(
-        [ '-d', '--date' ],
-        { help: 'Date to use as a threshold for removing old synthetic refs.[6 month default]' }
+        [ "-d", "--date" ],
+        { help: `Date to use as a threshold for removing old synthetic 
+            refs.[6 month default]` }
     );
 
     parser.addArgument(
-        [ '-f', '--force' ],
+        [ "-f", "--force" ],
         {
-            help: 'Actually remove synthetic refs. By Default we only list what refs we are about to remove.',
-            action: 'storeTrue',
+            help: `Actually remove synthetic refs. By Default we only list what 
+                refs we are about to remove.`,
+            action: "storeTrue",
             defaultValue: false,
         }
     );
@@ -44,17 +47,18 @@ function parseOptions() {
 function* removeSyntheticRef(repo, commit) {
     const refPath = SyntheticBranchUtil.getSyntheticBranchForCommit(commit);
 
-    console.log("Removing ref: " + refPath);
-
     if (SIMULATION) {
+        console.log("Removing ref: " + refPath);
         return;
     }
     
     return yield GitUtil.removeRemoteRef(repo, refPath);
 }
 
-// Go through the parents of 'commit' and remove synthetic reference recusively if they satisfy 'isDeletable'. 
-function* recursiveSyntheticRefRemoval(repo, commit, isDeletable, existingReferences) { 
+// Go through the parents of `commit` and remove synthetic reference recusively 
+// if they satisfy `isDeletable`. 
+function* recursiveSyntheticRefRemoval(repo, commit, isDeletable, 
+                                       existingReferences) { 
 
     assert.instanceOf(repo, NodeGit.Repository);
     assert.instanceOf(commit, NodeGit.Commit);
@@ -73,10 +77,25 @@ function* recursiveSyntheticRefRemoval(repo, commit, isDeletable, existingRefere
         if (isDeletable(parent) && existingReferences.includes(parent.sha())) {
             yield removeSyntheticRef(repo, parent);
         }
-        return yield *recursiveSyntheticRefRemoval(repo, parent, isDeletable, existingReferences);
+        return yield *recursiveSyntheticRefRemoval(repo, parent, isDeletable, 
+                                                   existingReferences);
     });
 
     return;
+}
+
+function* getSyntheticRefs(subRepo) {
+
+    const syntheticRefExtractor = function(value) {
+          if (value && value.includes(SYNTHETIC_BRANCH_BASE) ) {
+              return value.split("\t")[0];
+          }
+          return null;
+    };
+
+    let references = yield GitUtil.getRefs(subRepo);
+
+    return references.map(syntheticRefExtractor).filter(commit => commit);
 }
 
     // Clean up all redundant 
@@ -88,26 +107,13 @@ function* cleanUpRedundant(repo, roots, predicate) {
        let existingReferences = yield getSyntheticRefs(subRepo);
 
        for (let subCommit of roots[subName]) {
-           yield recursiveSyntheticRefRemoval(subRepo, subCommit, predicate, existingReferences); 
+           yield recursiveSyntheticRefRemoval(subRepo, subCommit, predicate, 
+                                              existingReferences); 
        }
    }
 }
 
-function* getSyntheticRefs(subRepo) {
-
-    const syntheticRefExtractor = function(value) {
-          if (value && value.includes(SYNTHETIC_BRANCH_BASE) ) {
-              return value.split("\t")[0];
-          }
-          return null;
-    }
-
-    let references = yield GitUtil.getRefs(subRepo);
-
-    return references.map(syntheticRefExtractor).filter(commit => commit);
-}
-
-    // Clean up all refs that satisfy 'isOldCommit', but not part of 'roots'.
+    // Clean up all refs that satisfy `isOldCommit`, but not part of `roots`.
 function* cleanUpOldRefs(repo, roots, isOldCommit) {
 
    for (let subName in roots) {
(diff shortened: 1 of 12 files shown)


node/lib/synthetic_gc.js
```diff
@@ -0,0 +1,231 @@
...
+    console.log("Removing ref: " + refPath);
+
+    if (SIMULATION) {
+        return;
+    }
+    
+    return yield GitUtil.removeRemoteRef(repo, refPath);
+}
+
+// Go through the parents of 'commit' and remove synthetic reference recusively if they satisfy 'isDeletable'.
```
novalis: replace with proper docstring as per the rest of git-meta

node/lib/synthetic_gc.js
```diff
@@ -0,0 +1,231 @@
...
+        if (isDeletable(parent) && existingReferences.includes(parent.sha())) {
+            yield removeSyntheticRef(repo, parent);
+        }
+        return yield *recursiveSyntheticRefRemoval(repo, parent, isDeletable, existingReferences);
+    });
+
+    return;
+}
+
+    // Clean up all redundant
```
novalis: "A ref is redundant if ..."

node/lib/util/git_util.js
```diff
@@ -453,6 +453,50 @@ exports.fetch = co.wrap(function *(repo, remoteName) {
...
 
+/**
+ * Fetch the remote having the specified the specified `repo`.
+ * Throw a `UserError` object if the repository cannot be fetched.
+ *
+ * @async
+ * @param {NodeGit.Repository} repo
+ */
+exports.getRefs = co.wrap(function *(repo) {
+
```
novalis: What's wrong with http://www.nodegit.org/api/reference/#list ? I don't think we should be using ls-remote anyway -- users should usually have the meta on the same machine as the subrepos.

node/lib/synthetic_gc.js
```diff
@@ -0,0 +1,231 @@
...
+       let refsToDelete = [];
+       for (let ref in references) {
+           // could not figure out how to call generators withing filter properly, so cant be fancy
+           const actualCommit = yield subRepo.getCommit(references[ref]);
+           if (isOldCommit(actualCommit)) {
+               refsToDelete.push(actualCommit);
+           }
+       }
+
+
```
novalis: nit: extra newline

node/lib/synthetic_gc.js
```diff
@@ -0,0 +1,231 @@
...
+       let references = yield getSyntheticRefs(subRepo);
+
+       // filter out all the references from rootA
+       references = references.filter(ref => !rawShaSubRoots.includes(ref)); 
+
+
+       // filter out all the references younger than 6 months
+       let refsToDelete = [];
+       for (let ref in references) {
+           // could not figure out how to call generators withing filter properly, so cant be fancy
```
novalis: Typically, you map to an array of callbacks, then call those callbacks in parallel by yielding the array, then you filter that.

node/lib/synthetic_gc.js
```diff
@@ -0,0 +1,231 @@
...
+
+function getThresholdDate(args) {
+
+    let date = new Date();
+
+    if (args.date != undefined) {
+        date = new Date(args.date);
+    } else {
+        const THRESHOLD_MONTHS = 6;
+        date.setMonth(date.getMonth() - THRESHOLD_MONTHS);
```
novalis: I cannot believe that this works, but apparently it does.  Yow.

node/lib/synthetic_gc.js
```diff
@@ -0,0 +1,231 @@
...
+    let date = new Date();
+
+    if (args.date != undefined) {
+        date = new Date(args.date);
+    } else {
+        const THRESHOLD_MONTHS = 6;
+        date.setMonth(date.getMonth() - THRESHOLD_MONTHS); 
+    }
+
+    console.log("Using: " + date + ", as threshold date.");
```
novalis: don't log.

node/lib/synthetic_gc.js
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// Fetch all important refs from out repo.
+function* populateRoots(repo) {
+
+    // For now, we are using heads/master as an important root.  
+    // 'important root' - means the root that most likey be around, so that we can remove all parent synthetic refs.
+    // This is not really necessary, more like optimization, 
+    // unless there is an important ref out there that was not updated for so long.
+    const IMPORTANT_REFS = ['refs/heads/fancy_branch', 'refs/heads/master'];
```
novalis: remove fancy_branch

node/lib/synthetic_gc.js
```diff
@@ -0,0 +1,231 @@
...
+    const submodules = yield SubmoduleUtil.getSubmoduleNames(repo);
+
+    const refs = yield repo.getReferenceNames(NodeGit.Reference.TYPE.LISTALL);
+    for (let ref in refs) {
+        ref = refs[ref]
+        const refHeadCommit = yield repo.getReferenceCommit(ref);
+
+        const tree = yield refHeadCommit.getTree();
+
+        // This could have been fancy if we were running gc per meta change
```
novalis: ?

node/lib/synthetic_gc.js
```diff
@@ -0,0 +1,231 @@
...
+            } else {
+                if (!(subName in classBroots)) {
+                    classBroots[subName] = new Set();
+                }
+                classBroots[subName].add(subCommit);
+            }
+        })
+    }
+
+    return classAroots;
```
novalis: what happens to classBroots?

node/lib/synthetic_gc.js
```diff
@@ -0,0 +1,231 @@
...
+    );
+    
+    return parser.parseArgs();
+}
+
+
+function* removeSyntheticRef(repo, commit) {
+    const refPath = SyntheticBranchUtil.getSyntheticBranchForCommit(commit);
+
+    console.log("Removing ref: " + refPath);
```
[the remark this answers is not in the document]
novalis: Oh, maybe if SIMULATION is true.

node/lib/util/git_util.js
zokir: Spoke with @bpeabody about this, he mentioned I should shell out for this. Otherwise, I will have problems with Kerberos.

node/lib/synthetic_gc.js
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    const parents = yield commit.getParents(commit.parentcount());
+    yield parents.map(function *(parent) {
+        if (parent in visited) {
+            return;
+        }
+        visited[parent] = 1;
+
+        if (isDeletable(parent) && existingReferences.includes(parent.sha())) {
```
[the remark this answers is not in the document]
zokir: Yes, idea was to try to delete synthetics refs that actually don't exist. Of course, if we always will regenerate all synthetics refs before running this. Then there is no use for this. 
I see it being helpful in simulation mode, where you would be albe to see what refs are actually being deleted within a particular run.

node/lib/util/git_util.js
novalis: Oh, right -- I forgot that we were using ls-remote.  But we don't want to use ls-remote -- we want to assume that all repos are on the same machine.

node/lib/synthetic_gc.js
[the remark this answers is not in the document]
novalis: Can you clarify this code -- by documenting what existingReferences is, and how it works?

node/lib/util/synthetic_gc_util.js
```diff
@@ -0,0 +1,330 @@
...
+SyntheticGcUtil.prototype.getBareSubmoduleRepo = co.wrap(
+    function*(repo, subName, refHeadCommit) {
+
+        assert.instanceOf(repo, NodeGit.Repository);
+        assert.instanceOf(refHeadCommit, NodeGit.Commit);
+
+        const fetcher = new SubmoduleFetcher(repo, refHeadCommit);
+
+        const subUrl = yield fetcher.getSubmoduleUrl(subName);
+        const subRepo = yield NodeGit.Repository.open(subUrl);
```
novalis: I think the URL will be "../../main/[something]", so this will only work if your cwd is correct, right?

node/lib/util/synthetic_gc_util.js
```diff
@@ -0,0 +1,330 @@
...
+    assert.instanceOf(repo, NodeGit.Repository);
+    assert.instanceOf(commit, NodeGit.Commit);
+
+    const synRefPath = SYNTHETIC_BRANCH_BASE + commit;
+
+    if (this.d_simulation) {
+        console.log("Removing ref: " + synRefPath);
+        return;
+    }
+    return NodeGit.Reference.remove(repo, synRefPath);
```
novalis: should we throw an exception if there's an error here?

node/lib/util/synthetic_gc_util.js
```diff
@@ -0,0 +1,330 @@
...
+ * @param {NodeGit.Commit}       commit
+ * @return {Number} 0 or an error code
+ */
+SyntheticGcUtil.prototype.removeSyntheticRef = co.wrap(
+    function(repo, commit) {
+
+    assert.instanceOf(repo, NodeGit.Repository);
+    assert.instanceOf(commit, NodeGit.Commit);
+
+    const synRefPath = SYNTHETIC_BRANCH_BASE + commit;
```
novalis: I think we already had a function somewhere that does this?

node/lib/util/synthetic_gc_util.js
```diff
@@ -0,0 +1,330 @@
...
+
+/**
+ * Go through the parents of `commit` of the specified `repo` and remove
+ * synthetic reference recursively if they satisfy `isDeletable` and not part of
+ * `existingReferences`.
+ *
+ * @param {NodeGit.Repository}   repo
+ * @param {NodeGit.Commit}       commit
+ * @param {Function}             isDeletable
+ * @param {String[]}             existingReferences
```
novalis: Is existingReferences really an array of strings, and if so do you really want to do a linear search of it on line 165?

node/lib/util/synthetic_gc_util.js
```diff
@@ -0,0 +1,330 @@
...
+    const parents = yield commit.getParents(commit.parentcount());
+    let thisState = this;
+    yield parents.map(function *(parent) {
+        if (parent.sha() in thisState.d_visited) {
+            return;
+        }
+        thisState.d_visited[parent.sha()] = 1;
+
+        // We are keeping track of 'existingReferences' to avoid trying to
+        // delete referneces that do not exists. This is helpful in simulation
```
novalis: spelling of referneces.

node/lib/util/synthetic_gc_util.js
```diff
@@ -0,0 +1,330 @@
...
+
+});
+
+/**
+ * Return all available synthetic refs within specified `subRepo`.
+ *
+ * @param {NodeGit.Repo}   subRepo
+ * @return {String[]}
+ */
+SyntheticGcUtil.prototype.getSyntheticRefs= co.wrap(
```
novalis: nit: space after =

node/lib/util/synthetic_gc_util.js
```diff
@@ -0,0 +1,330 @@
...
+    function*(subRepo) {
+
+    assert.instanceOf(subRepo, NodeGit.Repository);
+    let references = yield NodeGit.Reference.list(subRepo);
+
+    references = references.filter(
+            ref => ref.indexOf(SYNTHETIC_BRANCH_BASE) !== -1);
```
novalis: -== 0, not !== -1.

node/lib/util/synthetic_gc_util.js
```diff
@@ -0,0 +1,330 @@
...
+       }
+   }
+});
+
+/**
+ * Delete all synthetic refs within specified `repo` that satisfy `isOldCommit`,
+ * that not part of specified `roots`.
+ *
+ * @param {NodeGit.Repo}   repo
+ * @param {Object[]}       roots
```
novalis: I don't think roots is really an array -- I think it's a map from string to array, right?

node/lib/util/synthetic_gc_util.js
```diff
@@ -0,0 +1,330 @@
...
+
+   for (let subPath in roots) {
+
+       const subRepo = yield NodeGit.Repository.open(subPath);
+
+       const reservedCommits = roots[subPath];
+       const reservedSha = Array.from(reservedCommits).map(function(commit) {
+                            return commit.sha();
+                        });
+
```
novalis: nit: extra blank line

node/lib/util/synthetic_gc_util.js
```diff
@@ -0,0 +1,330 @@
...
+
+       let allRefs = yield this.getSyntheticRefs(subRepo);
+
+       for (let ref of allRefs) {
```
novalis: Whoa, for/of.  They finally unfucked for loops.

node/lib/util/synthetic_gc_util.js
```diff
@@ -0,0 +1,330 @@
...
+SyntheticGcUtil.prototype.populateRoots= co.wrap(
+    function*(repo) {
+
+    assert.instanceOf(repo, NodeGit.Repository);
+
+    // For now, we are using heads/master as an important root.
+    // `important root` - means the root that most likely be around, so that we
+    // can remove all parent synthetic refs.
+    // This is not really necessary, more like optimization,
+    // unless there is an important ref out there that was not updated for so
```
novalis: It seems likely that there would be such a ref.  Can we use refs/heads/*?

novalis: also refs/tags/*

node/lib/util/synthetic_gc_util.js
zokir: Url is absolute, tested in workdir and bare.
Ex:
subUrl: /Users/zokir/Documents/programming/node/work/remotes/subA.git/

node/lib/util/synthetic_gc_util.js
zokir: So, there this function in lib/synthetic_branch_util:60
`exports.getSyntheticBranchForCommit()`.

That I was calling before. But during the whole test ran it was very flaky by returning:
`return "refs/heads/metaTEST"`

I found a test that overrides this function in test/util/synthetic-branch.js:62
SyntheticBranch.getSyntheticBranchForCommit = function(commit) {                                                                 
            /*jshint unused:false*/                                                                                                      
         return "refs/heads/metaTEST";                                                                                                
};

I was not sure what is the proper way to fix it.
If you have any ideas I can submit pullrequest for it.

node/lib/util/synthetic_gc_util.js
novalis: In testing it is -- but what about in our actual repo, where the stored URL is relative?

node/lib/util/synthetic_gc_util.js
```diff
@@ -60,7 +62,11 @@ class SyntheticGcUtil {
     */
     constructor() {
         this.d_visited = {};
+        this.d_metaVisited = {};
         this.d_simulation = true;
+
+        // due to absense of value inequality in set
```
novalis: ?

node/lib/util/synthetic_gc_util.js
novalis: I removed that test override, so you shouldn't have that problem anymore.

node/lib/util/synthetic_gc_util.js
```diff
@@ -145,7 +159,7 @@ SyntheticGcUtil.prototype.removeSyntheticRef = co.wrap(
  * @param {NodeGit.Repository}   repo
  * @param {NodeGit.Commit}       commit
  * @param {Function}             isDeletable
- * @param {String[]}             existingReferences
+ * @param {Set<String>}          existingReferences
```
novalis: +@return

node/lib/util/synthetic_gc_util.js
```diff
@@ -272,59 +285,111 @@ SyntheticGcUtil.prototype.cleanUpOldRefs = co.wrap(
...
+* @return {Boolean}
+*/
+let isImportantRef = function(ref) {
+
+    // `important root` - means the root that most likely be around, so that we
+    // can remove all parent synthetic refs.
+    const IMPORTANT_REFS = ["refs/heads/", "refs/tags/"];
+
+    for (let importantRef of IMPORTANT_REFS) {
+        if (ref.indexOf(importantRef) === 0) {
```
novalis: if (ref.startsWith(importantRef)) {

node/test/util/synthetic_gc_util.js
```diff
@@ -271,7 +271,9 @@ describe("synthetic_gc_util", function () {
         yield syntheticGcUtil.cleanUpOldRefs(repo, roots, isOlderThanTomorrow);
 
         newSyntheticRefs = yield syntheticGcUtil.getSyntheticRefs(subRootRepo);
-        assert.equal(newSyntheticRefs.toString(), oid2.toString());
+        let EXPECTED_SYNTHETIC_REFS = 1;
```
novalis: nit: I don't love this variable naming.  First, all-caps vars are typically reserved for constants, not function-level vars.  Second, this variable holds the number of expected synthetic refs, not the refs themselves.  I would rather see let expectedSyntheticRefs = [oid2] and go from there.

node/lib/util/git_util.js
```diff
@@ -911,3 +911,66 @@ exports.configIsTrue = co.wrap(function*(repo, configVar) {
     return (configured === "true" || configured === "yes" ||
             configured === "on");
 });
+
+/**
+ * Fetch the remote having the specified `repo`.
```
novalis: "Fetch and return a list of remote refs in the origin for the specified `repo` "

node/lib/util/git_util.js
```diff
@@ -911,3 +911,66 @@ exports.configIsTrue = co.wrap(function*(repo, configVar) {
...
+ * Throw a `UserError` object if the repository cannot be fetched.
+ *
+ * @async
+ * @param {NodeGit.Repository} repo
+ */
+exports.getRefs = co.wrap(function *(repo) {
+	assert.instanceOf(repo, NodeGit.Repository);
+
+    const execString = `git -C '${repo.path()}' ls-remote --refs`;
```
novalis: "... -- refs/" instead "... -- refs" (that is, add a trailing slash)

node/lib/util/git_util.js
```diff
@@ -911,3 +911,66 @@ exports.configIsTrue = co.wrap(function*(repo, configVar) {
...
+
+/**
+ * Delete the specified `ref` on the specified remote `repo`.
+ * Throw a `UserError` object if cannot push to remote repository.
+ *
+ * @async
+ * @param {NodeGit.Repository} repo
+ * @param {String}             remoteName
+ */
+exports.removeRemoteRef = co.wrap(function *(repo, ref) {
```
novalis: Given the likely number of these, I wonder if it would make sense to do batch deletion.  That is, push multiple deletes at once.

node/lib/util/synthetic_gc_util.js
```diff
@@ -37,13 +37,57 @@ const NodeGit = require("nodegit");
...
+    },
+
+    /**
+     * Remove gubbins from the specified 'path'.
+     *
+     * @param {String}      path
+     * @return {String}
+     */
+    rebuildUrl : function(path) {
+        path = path.replace(new RegExp("../main/"), "");
```
novalis: This stuff needs to be configurable.  Also, dots in a regex are metacharacters; if you want literal dots, you need to escape them.

node/lib/util/synthetic_gc_util.js
```diff
@@ -37,13 +37,57 @@ const NodeGit = require("nodegit");
...
+     *
+     * @param {NodeGit.Repository} repo
+     * @return {Set<String>}
+     */
+    getLocalSyntheticRefs : function*(repo) {
+        assert.instanceOf(repo, NodeGit.Repository);
+
+        let references = yield NodeGit.Reference.list(repo);
+        references = references.filter(
+                ref => ref.indexOf(SYNTHETIC_BRANCH_BASE) === 0);
```
novalis: startsWith

node/lib/util/synthetic_gc_util.js
```diff
@@ -37,13 +37,57 @@ const NodeGit = require("nodegit");
...
+    rebuildUrl : function(path) {
+        path = path.replace(new RegExp("../main/"), "");
+        path = path.replace(new RegExp(".d/", "g"), "/");
+        if (path.endsWith("-git")) {
+            path = path.replace(new RegExp("-git$"), ".git");
+        }
+    },
+
+    /**
+     * Return a collection of synthetic refs reachable locally for specified
```
novalis: s/reachable//

node/lib/util/synthetic_gc_util.js
```diff
@@ -112,14 +156,29 @@ SyntheticGcUtil.prototype.getBareSubmoduleRepo = co.wrap(
...
-            subUrl = SubmoduleConfigUtil.resolveSubmoduleUrl(metaUrl,
-                                                             subUrl);
+        if (subUrl.startsWith(".")) {
+            // relative url specified
+            subUrl = detail.rebuildUrl(subUrl);
+            subPath = yield SyntheticBranchUtil.urlToLocalPath(repo, subUrl);
+        }
+
+        if (!fs.existsSync(subPath)) {
+            console.log("Submodule at following path does not exists: " +
```
novalis: s/exists/exist/ -- and do we want to output this?  It seems like the sort of thing we would want only for "--verbose"

node/lib/util/synthetic_gc_util.js
```diff
@@ -112,14 +156,29 @@ SyntheticGcUtil.prototype.getBareSubmoduleRepo = co.wrap(
...
+                subPath);
+
+            const metaUrl = yield GitUtil.getOriginUrl(repo);
+            if (metaUrl !== null) {
```
novalis: When would this be null, and do we really want to not fetch in that case?  Should we just fail loudly here?

node/lib/util/synthetic_gc_util.js
```diff
@@ -37,13 +37,57 @@ const NodeGit = require("nodegit");
...
-    return existingReferences.has(sha);
+let detail = {
+    /**
+     * Return whether sha is in existingReferences.
+     *
+     * @param {Set<String>} existingReferences
+     * @param {String}      sha
+     * @return {Boolean}
+     */
+    refExists : function(existingReferences, sha) {
```
novalis: I think we can just inline this everywhere it's used -- having it here doesn't really save anything right?

node/lib/util/synthetic_gc_util.js
```diff
@@ -202,15 +263,30 @@ SyntheticGcUtil.prototype.recursiveSyntheticRefRemoval = co.wrap(
...
-            ref => ref.indexOf(SYNTHETIC_BRANCH_BASE) === 0);
+    // first lets check if we are on the git server itself.
+    let references = yield detail.getLocalSyntheticRefs(repo);
+    if (references.size !== 0) {
+        return references;
+    }
 
-    references = references.map(ref => ref.split(SYNTHETIC_BRANCH_BASE)[1]);
+    const syntheticRefExtractor = function(value) {
+        if (value && value.includes(SYNTHETIC_BRANCH_BASE)) {
```
novalis: s/includes/startsWith/

node/lib/util/git_util.js
novalis: also maybe rename this to getRemoteRefs?

node/lib/util/synthetic_gc_util.js
```diff
@@ -202,15 +263,30 @@ SyntheticGcUtil.prototype.recursiveSyntheticRefRemoval = co.wrap(
...
+            return value.split("\t")[0];
+        }
+        return null;
+    };
+
+    references = yield GitUtil.getRefs(repo);
+
+    if (!Array.isArray(references)) {
```
novalis: This should never happen, right?

node/lib/util/synthetic_gc_util.js
```diff
@@ -422,12 +424,13 @@ SyntheticGcUtil.prototype.populatePerCommit = co.wrap(
                 classAroots[subPath].add(subCommit);
                 this.d_subCommitStored[subPath][subCommit.sha()] = 1;
             } catch(exception) {
-                console.log("Cannot process submodule with following " +
-                    " exception: " + exception);
+                console.log("Cannot process submodule " + subName +
```
novalis: we should probably at the very least:
1. log this to stderr and
2. exit with a non-zero status eventually, even if we do not do so immediately.

node/lib/util/synthetic_gc_util.js
```diff
@@ -68,6 +68,8 @@ let detail = {
         if (path.endsWith("-git")) {
             path = path.replace(new RegExp("-git$"), ".git");
         }
+
+        return path;
```
novalis: Did the tests pass before this fix? If so, do we need to change them?

node/lib/util/git_util.js
zokir: `"... -- refs/" instead "... -- refs" (that is, add a trailing slash)`

Not sure what you mean by this, why  -- refs /? ls-remote takes --refs flag.

novalis: Oops, I totally misread this.  Nevermind.

node/lib/synthetic_gc.js
```diff
@@ -145,6 +145,23 @@ function getThresholdDate(args) {
     return date;
 }
 
+/**
+ * Set 'gitmeta.subreporootpath' if its not set.
+ *
+ */
+let setRepoRootPathIfNotConfigured = function*(repo) {
```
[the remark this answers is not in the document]
zokir: This is needed by synthetic_branch_util:urlToLocalPath.

Basically, if you have a submodule with the relative path you need to set it.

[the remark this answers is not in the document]
zokir: I wrote a test that demonstrates it, will show in next pr.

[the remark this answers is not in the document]
novalis: I mean, I know why that variable would be useful.  I don't understand why it's OK to set it for the user.  It seems to me that only the user should know the correct value.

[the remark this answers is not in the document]
zokir: ok, will remove

node/lib/util/synthetic_gc_util.js
zokir: Yeah, tests were using absolute paths for submodules. So this code was not hit, added test with relative paths. Thanks for catching.

node/lib/synthetic_gc.js
```diff
@@ -0,0 +1,191 @@
...
+                refs we are about to remove.`,
+            action: "storeTrue",
+            defaultValue: false,
+        }
+    );
+
+    return parser.parseArgs();
+}
+
+let lessThanDate = function(thresHold) {
```
bpeabody: function lessThanDate(...

node/lib/synthetic_gc.js
```diff
@@ -0,0 +1,191 @@
...
+    try {
+        yield config.getStringBuf("gitmeta.subreporootpath");
+    } catch (exception) {
+        console.log("cannot get subrepopath with following exception " +
+            exception + ", setting to current directory");
+        yield config.setString("gitmeta.subreporootpath", process.cwd());
+    }
+};
+
+let runIt = co.wrap(function *(args) {
```
bpeabody: `const runIt...`

node/test/util/synthetic_gc_util.js
```diff
@@ -0,0 +1,321 @@
...
+    const result = yield SubmoduleConfigUtil.initSubmoduleAndRepo(
+                                                             originUrl,
+                                                             repo,
+                                                             subName,
+                                                             url,
+                                                             null);
+    return result;
+});
+
+let isDeletable = function() {
```
bpeabody: `function isDeletable`

node/lib/util/synthetic_gc_util.js
novalis: I still don't understand this comment

node/lib/util/synthetic_gc_util.js
```diff
@@ -167,17 +172,26 @@ SyntheticGcUtil.prototype.getBareSubmoduleRepo = co.wrap(
...
 
             const metaUrl = yield GitUtil.getOriginUrl(repo);
-            if (metaUrl !== null) {
-                subUrl = yield fetcher.getSubmoduleUrl(subName);
-                subUrl = SubmoduleConfigUtil.resolveSubmoduleUrl(metaUrl,
-                                                                 subUrl);
+            if (metaUrl === null) {
+                console.error("Cannot determine origin url. Something is " +
+                    "wrong, exiting.");
+                process.exit(-1);
```
novalis: should we throw an exception?

node/lib/util/git_util.js
```diff
@@ -911,3 +911,71 @@ exports.configIsTrue = co.wrap(function*(repo, configVar) {
...
+	assert.instanceOf(repo, NodeGit.Repository);
+
+    const execString = `git -C '${repo.path()}' ls-remote --refs`;
+
+    try {
+        const result = yield ChildProcess.exec(execString);
+        return result.stdout.split("\n");
+    }
+    catch (e) {
+        return new UserError(e.message);
```
bpeabody: `throw new UserError`

node/lib/util/git_util.js
```diff
@@ -911,3 +911,71 @@ exports.configIsTrue = co.wrap(function*(repo, configVar) {
...
+        return;
+    }
+
+    const execString = `git -C '${repo.path()}' push origin :` +
+                        refs.join(" :");
+    try {
+        return yield ChildProcess.exec(execString);
+    }
+    catch (e) {
+        return new UserError(e.message);
```
bpeabody: and here

node/lib/util/synthetic_gc_util.js
```diff
@@ -0,0 +1,541 @@
...
+const SubmoduleConfigUtil = require("./submodule_config_util");
+const SubmoduleFetcher = require("./submodule_fetcher");
+const SyntheticBranchUtil = require("./synthetic_branch_util");
+const UserError = require("./user_error");
+const GitUtil = require("./git_util");
+const fs = require("fs");
+
+const SYNTHETIC_BRANCH_BASE = "refs/commits/";
+
+let detail = {
```
bpeabody: `const detail =`

novalis:
The timeout is needed because we push synthetic refs before real branches, and depend on them for referential integrity. This logic is the same as git's logic for object GC.

On January 16, 2018 9:42:57 AM EST, Brock Peabody <notifications@github.com> wrote:
>bpeabody requested changes on this pull request.
>
>I would like to make two simplifications:
>
>1. Remove the concept of "important" refs. We don't know what is
>important at the level of this tool. If we do say that, e.g.,
>`refs/foo` is not-important, then this algorithm puts us in a state
>where `refs/foo` may no longer be invalid -- rather than impacting the
>GC logic, we may as well remove `refs/foo`. However, if we do determine
>(by policy) that "non-important" refs no longer need to be maintained
>after some time, we may as well just delete those refs and let "nature
>take its course" -- the process for cleaning up old unimportant refs
>would not be part of this GC tool.
>2. Remove any time consideration at all. If a synthetic ref is
>redundant, we can remove it immediately.
>
>
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/twosigma/git-meta/pull/445#pullrequestreview-89118202

-- 

bpeabody:
@novalis : that makes sense re. the timeout. Do you agree with my point about the "important" refs?

novalis:
I think so, but run it by Adam.

On January 16, 2018 10:20:37 AM EST, Brock Peabody <notifications@github.com> wrote:
>@novalis : that makes sense re. the timeout. Do you agree with my point
>about the "important" refs?
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/twosigma/git-meta/pull/445#issuecomment-357993798

-- 

node/lib/util/synthetic_gc_runner.js
```diff
@@ -0,0 +1,473 @@
...
+const SubmoduleConfigUtil = require("./submodule_config_util");
+const SubmoduleFetcher = require("./submodule_fetcher");
+const SyntheticBranchUtil = require("./synthetic_branch_util");
+const UserError = require("./user_error");
+const GitUtil = require("./git_util");
+const fs = require("fs");
+
+const SYNTHETIC_BRANCH_BASE = "refs/commits/";
+
+const detail = {
```
bpeabody: You need a `detail` "namespace" here. These names are hidden in this file unless explicitly exported.

node/lib/util/synthetic_gc_runner.js
```diff
@@ -0,0 +1,473 @@
...
+ * within specified meta `repo`.
+ *
+ * @param {NodeGit.Repository}   repo
+ * @param {NodeGit.Commit}       commit
+ * @return {NodeGit.Repository}  subRepo
+ */
+SyntheticGcRunner.prototype.getBareSubmoduleRepo = co.wrap(
+    function*(repo, subName, refHeadCommit) {
+
+        assert.instanceOf(repo, NodeGit.Repository);
```
bpeabody: 4-space indentation is enough here

node/lib/util/synthetic_gc_runner.js
```diff
@@ -0,0 +1,473 @@
...
+/**
+ * Return bare submodule repository corresponding to a specified `refHeadCommit`
+ * within specified meta `repo`.
+ *
+ * @param {NodeGit.Repository}   repo
+ * @param {NodeGit.Commit}       commit
+ * @return {NodeGit.Repository}  subRepo
+ */
+SyntheticGcRunner.prototype.getBareSubmoduleRepo = co.wrap(
+    function*(repo, subName, refHeadCommit) {
```
bpeabody: I'd add assertions for the types of all the arguments.

node/lib/util/synthetic_gc_runner.js
```diff
@@ -0,0 +1,473 @@
...
+ * @param {NodeGit.Commit}       commit
+ */
+SyntheticGcRunner.prototype.removeSyntheticRef = function(repo, commit) {
+
+    assert.instanceOf(repo, NodeGit.Repository);
+    assert.instanceOf(commit, NodeGit.Commit);
+
+    const synRefPath = SyntheticBranchUtil.getSyntheticBranchForCommit(commit);
+
+    if (this.d_simulation) {
```
bpeabody: This seems a little odd to me; I'd rather not call this method when simulating than to have it check for simulation.

node/lib/util/synthetic_gc_runner.js
```diff
@@ -0,0 +1,473 @@
...
+
+    assert.instanceOf(repo, NodeGit.Repository);
+
+    // first lets check if we are on the git server itself.
+    let references = yield detail.getLocalSyntheticRefs(repo);
+    if (references.size !== 0) {
+        return references;
+    }
+
+    const syntheticRefExtractor = function(value) {
```
bpeabody: space after `function`

node/lib/util/synthetic_gc_runner.js
```diff
@@ -0,0 +1,473 @@
...
+ *
+ * @param {NodeGit.Repo}   repo
+ * @param {NodeGit.Commit} commit
+ * @param {Map<String, Set<String>>} classAroots
+ * @return {Map<String, Set<String>>}
+ */
+SyntheticGcRunner.prototype.populatePerCommit = co.wrap(
+    function*(repo, commit, classAroots) {
+
+        if (commit.sha() in this.d_metaVisited) {
```
bpeabody: again, just 4 space indentation

node/lib/util/synthetic_gc_runner.js
```diff
@@ -0,0 +1,473 @@
...
+                    this.d_subCommitStored[subPath] = {};
+                }
+
+                if (subCommit.sha() in this.d_subCommitStored[subPath]) {
+                    continue;
+                }
+
+                classAroots[subPath].add(subCommit);
+                this.d_subCommitStored[subPath][subCommit.sha()] = 1;
+            } catch(exception) {
```
bpeabody: `catch` next line

PR #449, opened 2017-12-14
make synthetic referential integrity check work on shas not refs

novalis:


node/lib/util/shorthand_parser_util.js
```diff
@@ -1136,10 +1136,11 @@ exports.parseMultiRepoShorthand = function (shorthand, existingRepos) {
     function addCommit(id, commit) {
         if (id in commits) {
             const oldCommit = commits[id];
+            const msg = `different submodule commit for meta commit ${id}`;
```
novalis: I was very confused by the original message (and not just because of how different was spelled).  Does the new message seem correct to you?

PR #454, opened 2017-12-26
Forward all non-meta commands to Git.

bpeabody:
Instead of forwarding just a whitelist.

addresses https://github.com/twosigma/git-meta/issues/451

### DIFF
--- a/node/lib/git-meta.js
+++ b/node/lib/git-meta.js
@@ -109,7 +109,9 @@ commands will generally perform that same operation, but across a *meta*
 repository and the *sub* repositories that are locally *opened*.  These
 commands work on any Git repository (even one without configured submodules);
 we do not provide duplicate commands for Git functionality that does not need
-to be applied across sub-modules such as 'clone' and 'init'.`;
+to be applied across sub-modules such as 'clone' and 'init'.  Note that
+git-meta will forward any subcommand that it does not implement to Git,
+as if run with 'git -C $(git meta root) ...'.`;
 
 const parser = new ArgumentParser({
     addHelp:true,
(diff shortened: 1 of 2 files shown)


node/lib/git-meta.js
```diff
@@ -160,7 +156,9 @@ Object.keys(commands).sort().forEach(name => {
 // sub-parser level.
 
 if (2 < process.argv.length &&
-    Forward.forwardedCommands.has(process.argv[2])) {
+    "-h" !== process.argv[2] &&
```
novalis: --version?

novalis:
Others we might want to skip for now are:
- `am` -- part of the rebase machinery
- `annotate` aka `blame` -- doesn't understand submodules
- `archimport`, `cvs\*`, 'svn', 'p4' -- these don't make sense for a meta repo
- `archive` -- doesn't work as user would expect in a meta repo
- `clean`, `filter-branch`, `grep` -- these don't  recurse the way a user would expect
- `fast-import`, `fast-import` -- users should not be using these without understanding what they are doing.
- `merge-*` -- these are subcommands of merge that I think we don't expect users to use directly
- `mv`, `quiltimport`, `rm` -- we would need to implement these inside of meta, like `revert`
- `shell` -- users should never be using this anyway
- `stage` -- this is an alias for add that we should implement internally
- `worktree` -- this should print an error message explaining that git-meta doesn't work with worktrees due to libgit2 badness.

Some of these are obviously more important than others.  `mv` and `rm` are probably the most important after `revert`.

bpeabody:
@novalis: I changed to an explicit blacklist and added all the commands you suggested.

PR #468, opened 2018-02-02
support push-to-url

novalis:


node/test/util/git_util.js
```diff
@@ -199,6 +199,18 @@ describe("GitUtil", function () {
         });
     });
 
+    describe("getUrlFromRemoteName", function () {
+        it("works", co.wrap(function *() {
+            const repo = yield TestUtil.createSimpleRepository();
+            const func = GitUtil.getUrlFromRemoteName;
+            yield NodeGit.Remote.create(repo, "upstream",
+                                        "https://example.com");
+            assert.equal("https://example.com", yield func(repo, "upstream"));
```
bpeabody: Maybe this is a bit nitty, but ideally you'd want to exercise the "@" and "http:" cases too. This is probably good enough for now though. I'll approve and you can add those cases if you wish.

novalis: In theory, yes.  In practice, there's no strong reason to do it -- the code is pretty likely to be correct, since it's simple and transparent.

bpeabody: Fine with me.

Issue #183, opened 2017-01-23
`git meta status` should not show relative path for submodule name


bpeabody:
E.g., if there is a new commit in `s`:

```bash
$ cd s
$ git meta status
modified:      (submodule, on unrelated commit)
```

bpeabody:
I will change this to print the full submodule name, prefixed with a `/` so that it's clear it's not a relative path like other paths listed in `status`.

PR #478, opened 2018-02-10
Stop showing submodule status as relative

bpeabody:
This causes the name to appear as blank when you're in the submodule,
and is not generally useful anyway.

Addresses: https://github.com/twosigma/git-meta/issues/183


novalis:
I actually liked relative -- it's what git commands tend to do.  I just prefer '.' to the empty string.  Wouldn't that be better?

bpeabody:
I'm OK with that.

Issue #484, opened 2018-02-14
honor `branch.<name>.pushRemote` and `remote.pushDefault`


bpeabody:
As described: https://git-scm.com/docs/git-config

PR #485, opened 2018-02-14
Honor `pushDefault` and `pushRemote` in `push`

bpeabody:
Addresses https://github.com/twosigma/git-meta/issues/484


node/lib/util/git_util.js
```diff
@@ -131,16 +134,42 @@ exports.getTrackingInfo = co.wrap(function *(branch) {
...
+
+    // Try to read a 'pushRemote' for the branch.
+
+    let pushRemote = null;
+    try {
+        pushRemote = yield config.getStringBuf(
+                                    `branch.${branch.shorthand()}.pushRemote`);
+    }
+    catch (e) {
+        // Unfortunately, no other way to handle a missing config entry
```
novalis: I feel like we do this a lot -- could we extract a method?

bpeabody: OK, see if this is better.

Issue #495, opened 2018-02-28
Add git-meta hooks.


s-xiang:
Add missing hooks to git-meta:
* checkout
* rebase
* merge 
* open
* close
* add-submodule
*  reset
* cherry-pick

PR #496, opened 2018-02-28
#495: add missing hooks in git-meta.

s-xiang:


node/lib/cmd/rebase.js
```diff
@@ -110,4 +111,7 @@ exports.executeableSubcommand = co.wrap(function *(args) {
         const commit = yield repo.getCommit(committish.id());
         yield RebaseUtil.rebase(repo, commit);
     }
+
+    // Run post-rewrite hook with "rebase" as arg, means rebase command invoked this hook.
+    yield Hook.execHook("post-rewrite", args=["rebase"]);
```
novalis: s/args=//

novalis: (it's not doing what you think it's doing)

node/lib/util/hook.js
```diff
@@ -0,0 +1,59 @@
...
+ */
+exports.execHook = co.wrap(function *(name, args=[]) {
+    assert.isString(name);
+
+    const hookPath = path.join(GitUtil.getRootGitDirectory(), ".git/hooks");
+    const absPath = path.resolve(hookPath, name);
+
+    try {
+        const result = yield ChildProcess.execFile(absPath, args);
+        return result.stdout;
```
bpeabody: If this method has a return value, please document it above.

bpeabody: Explicitly call out the arguments in the documentation and list their names in the jsduck like:
```
@param {String} name ...
@param {String[]} args ...
```

Issue #509, opened 2018-03-12
refuse to run `meta reset --hard $SHA` if $SHA does not contain a currently open submodule with unstaged changes


bpeabody:
Due to https://github.com/twosigma/git-meta/issues/508 (for which this issue is a workaround), the submodule directory would be deleted and these changes lost for good.

PR #510, opened 2018-03-12
Stop `reset --hard` if will remove dirty open subs

bpeabody:
In this case, libgit2 will delete the submodule's directory, throwing
away all changes.  See https://github.com/twosigma/git-meta/issues/509

I do not want to do this check when resetting to abort cherry-pick -- we
started out clean before picking and it could be pretty common that
we've added a submodule that we want to get rid of when we abort -- but
I don't want to add a weird flag that will exist just for this
workaround.  So, I went ahead and added a `MERGE` reset type, that will
eventually do a real `git reset --merge` but for the time being is just
like `HARD` except that it does not do this check.

### DIFF
--- a/node/lib/util/reset.js
+++ b/node/lib/util/reset.js
@@ -32,6 +32,7 @@
 
 const assert  = require("chai").assert;
 const co      = require("co");
+const colors  = require("colors");
 const NodeGit = require("nodegit");
 
 const DoWorkQueue         = require("../util/do_work_queue");
@@ -46,6 +47,7 @@ const TYPE = {
     SOFT: "soft",
     MIXED: "mixed",
     HARD: "hard",
+    MERGE: "merge",
 };
 Object.freeze(TYPE);
 exports.TYPE = TYPE;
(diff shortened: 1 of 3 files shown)


node/lib/util/reset.js
```diff
@@ -60,10 +62,59 @@ function getType(type) {
         case TYPE.SOFT : return NodeGit.Reset.TYPE.SOFT;
         case TYPE.MIXED: return NodeGit.Reset.TYPE.MIXED;
         case TYPE.HARD : return NodeGit.Reset.TYPE.HARD;
+
+        // TODO: real implementation of `reset --merge`.  For now, this behaves
```
novalis: maybe instead just throw an exception?  I've never heard of anyone using --merge.

bpeabody: You don't use it directly, but it's what the rebase/merge/cherry-pick `--abort` commands use to restore your repo.

```
           --merge
               Resets the index and updates the files in the working tree that are different between <commit> and HEAD, but keeps those which are
               different between the index and working tree (i.e. which have changes which have not been added). If a file that is different between
               <commit> and the index has unstaged changes, reset is aborted.
```

PR #511, opened 2018-03-14
Support for reading and writing sequencer states

bpeabody:
Addreses task: https://github.com/twosigma/git-meta/projects/2#card-8117068

### DIFF
--- a/node/lib/util/sequencer_state.js
+++ b/node/lib/util/sequencer_state.js
@@ -1,0 +1,210 @@
+/*
+ * Copyright (c) 2018, Two Sigma Open Source
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of git-meta nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+"use strict";
+
+const assert  = require("chai").assert;
+const deeper   = require("deeper");
+
+const TYPE = {
+    CHERRY_PICK: "CHERRY_PICK",
+    MERGE: "MERGE",
+    REBASE: "REBASE",
+};
+
+/**
+ * This module defines the `SequencerState` value-semantic type.
+ */
+
+/**
+ * @class CommitAndRef
+ *
+ * This class describes a commit and optionally the ref it came from.
+ */
+class CommitAndRef {
+    /**
+     * Create a new `CommitAndRef` object.
+     *
+     * @param {String}      sha
+     * @param {String|null} ref
+     */
+    constructor(sha, ref) {
+        assert.isString(sha);
+        if (null !== ref) {
+            assert.isString(ref);
+        }
+        this.d_sha = sha;
+        this.d_ref = ref;
+
+        Object.freeze(this);
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


node/lib/util/sequencer_state.js
```diff
@@ -0,0 +1,156 @@
...
+};
+
+/**
+ * This module defines the `SequencerState` value-semantic type.
+ */
+
+/**
+ * @class CommitAndRef
+ *
+ * This class describes a commit and optionally the ref it came from.
```
novalis: Isn't that just a NodeGit.AnnotatedCommit?  Maybe there's no way to get the ref out of that, tho.

node/lib/util/sequencer_state_util.js
```diff
@@ -0,0 +1,237 @@
...
+ * @return {CommitAndRef|null}
+ */
+exports.readCommitAndRef = co.wrap(function *(gitDir, fileName) {
+    assert.isString(gitDir);
+    assert.isString(fileName);
+    const content = yield exports.readFile(gitDir, fileName);
+    if (null !== content) {
+        const lines = content.split("\n");
+
+        // always an extra entry after the split
```
novalis: Do you mean that the file consists of x\ny\n, and so split gives ['x', 'y', '']?  If so, strip the trailing newline before split to make clearer what's going on.

node/lib/util/sequencer_state_util.js
```diff
@@ -0,0 +1,237 @@
...
+        if (1 === numLines || 2 === numLines) {
+            const ref = 2 === numLines ? lines[1] : null;
+            return new CommitAndRef(lines[0], ref);
+        }
+    }
+    return null;
+});
+
+/**
+ * Return the array of commit stored in the specified `gitDir`, or null if
```
novalis: Can you describe a bit better what this file is?  All I know now is that it's a list of commits -- but what for?

novalis: Actually, maybe describe the directory schema at the top of the file, because I have no idea why we are reading these files or whatever.

node/lib/util/sequencer_state_util.js
```diff
@@ -0,0 +1,237 @@
...
+/**
+ * Return the index of the current commit stored in the specified `gitDir`
+ * having the specified `numCommits` number of commits  or null if the file is
+ * missing, malformed, our out-of-bounds.
+ *
+ * @param {String} gitDir
+ * @param {Number} numCommits
+ * @return {Number}
+ */
+exports.readCurrentCommit = co.wrap(function *(gitDir, numCommits) {
```
novalis: I don't understand what numCommits means or what values are valid.

node/lib/util/sequencer_state_util.js
```diff
@@ -0,0 +1,237 @@
...
+            }
+        }
+    }
+    return null;
+});
+
+/**
+ * Return the sequencer state if it exists in the specified `gitDir`, or null
+ * if it is missing or malformed.
+ * TODO: emit diagnostic when malformed?
```
novalis: raise exception if malformed -- let the caller say, "hey, your sequencer state is malformed.  This shouldn't happen.  You can fix it by deleting .git/meta-sequencer and then using the reflog to get back to where you started."

(ok to leave like this for this pr)

node/lib/util/sequencer_state_util.js
```diff
@@ -0,0 +1,237 @@
...
+    const commits = yield exports.readCommits(gitDir);
+    if (null === commits) {
+        return null;                                                  // RETURN
+    }
+    const currentCommit = yield exports.readCurrentCommit(gitDir,
+                                                          commits.length);
+    if (null === currentCommit) {
+        return null;                                                  // RETURN
+    }
+    return new SequencerState(type, original, target, commits, currentCommit);
```
novalis: Would a fluent builder pattern be better here, or is that not a thing in JS?

node/lib/util/sequencer_state.js
bpeabody: Yes, that's what I'd use ideally, but there's no way to make an `AnnotatedCommit` without a repo (which makes it harder to test with them, and as you point out, no way to get the ref back.

node/lib/util/sequencer_state_util.js
bpeabody: good idea

node/lib/util/sequencer_state_util.js
bpeabody: Yes, I want to think about this some more.

node/lib/util/sequencer_state_util.js
bpeabody: To deal reduce the number of args? I think what you'd be more likely to see in JS would be a dictionary, especially if you have a optional arguments. I didn't use one because all the args here are required, but it would look like:
```JavaScript
SequencerState({
    type: type
    original: original,
    target: target,
    commits: commits,
    currentCommit: currentCommit,
});
```

bpeabody:
OK, let's see if this version is more comprehensible.

node/lib/util/sequencer_state_util.js
novalis: Not just reduce the number, but also avoid having to look up ordering.  I like that version better.

bpeabody: I'm fine with that. I could see this constructor getting more args in the future, and I don't think I'd go to 6.

node/lib/util/sequencer_state_util.js
```diff
@@ -0,0 +1,254 @@
...
+
+/**
+ * This module contains methods for accessing, reading, and rendering
+ * `SequencerState` objects on disk.  The disk structure is (note, all files
+ * are eol-terminated):
+ * 
+ * .git/meta_sequencer/TYPE            -- Single line containing the type, e.g.
+ *                                        "REBASE".
+ *                    /ORIGINAL_HEAD   -- SHA and optional ref name of what was
+ *                                     -- on head.  If first line is the SHA.
```
novalis: sentence fragment "If the first line is the SHA" is not a syntactic constituent.

bpeabody:
OK, I think I addressed all the concerns, but I also slipped in a couple of methods that I will need but had forgotten: `equal` and `toString` if you want to take a quick look at them, @novalis .

node/lib/util/sequencer_state.js
```diff
@@ -0,0 +1,210 @@
...
+
+    /**
+     * Return true if the specified `rhs` represents the same value as this
+     * object.  Two `CommitAndRef` objects represet the same value if they have
+     * the same `sha` and `ref` properties.
+     *
+     * @param {CommitAndRef} rhs
+     * @return {Bool}
+     */
+    equal(rhs) {
```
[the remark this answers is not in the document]
bpeabody: I think the types they're referring to in that doc (e.g., `Hash`) are a library that's not part of standard JS, which, unfortunately doesn't seem to have standard operators that I've seen for equality testing and what not. In any case, I made an `equal`, not an `equals`.

Either way, I've been using `equal` in other classes. If this code is OK otherwise, I'd be happy to go back and revisit across the board later.

novalis:
Oy, Javascript.

PR #513, opened 2018-03-23
Change to root git directory before executing hooks

zokir:
git switches to git root directory before executing hooks.
https://git-scm.com/docs/githooks#_description

We should do the same to avoid problems. (you know the use case)


zokir:
@novalis

PR #515, opened 2018-03-23
Sequencer testing

bpeabody:
See card: https://github.com/twosigma/git-meta/projects/2#card-8117137

### DIFF
--- a/node/lib/util/shorthand_parser_util.js
+++ b/node/lib/util/shorthand_parser_util.js
@@ -31,8 +31,11 @@
 "use strict";
 
 const assert      = require("chai").assert;
+
 const RepoAST     = require("../util/repo_ast");
 const RepoASTUtil  = require("../util/repo_ast_util");
+
+const SequencerState = RepoAST.SequencerState;
 
 /**
  * @module {ShorthandParserUtil}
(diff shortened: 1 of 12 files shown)


node/lib/util/shorthand_parser_util.js
```diff
@@ -79,6 +83,10 @@ const RepoASTUtil  = require("../util/repo_ast_util");
  * rebase         = E<head name>,<original commit id>,<onto commit id>
  * merge          = M<message>,<original head>,<merge head>
  * cherry-pick    = P<original head>,<picked commit>
+ * sequencer      = Q<sequencer type>' '<commit and ref>' '<commit and ref>' '
```
novalis: Can you describe what the format of commit-and-ref is as well as the semantics of each value?  As in &lt;original-head-ref:sha&gt;

bpeabody: It's just a couple of lines under that one:

```
 * sequencer      = Q<sequencer type>' '<commit and ref>' '<commit and ref>' '
 *                  <number>' '<commit>(','<commit>)*
 * sequencer type = C | M | R
 * commit and ref = commit':'[<name>]
```

bpeabody: I've got an example a little further down too:

```
 * QR 1:refs/heads/master 8: 1 3,5,1
 *                              -- A sequencer is in progress that is a 
 *                              -- rebase.  When the rebase started, HEAD
 *                              -- was on `master`.  
```

novalis: Oh, the format is -- that makes sense. But the semantics are still obscure to me.

PR #529, opened 2018-04-06
Upgrade nodegit -- work-in-progress, do not merge

novalis:
While I've been working on git meta rm, I noticed at least two bugs in nodegit which were fixed in more-recent versions.  This upgrade passes all tests except the rebase tests, and since Brock is redoing rebase anyway, I figured I would post this for discussion.  @bpeabody what do you think?  Keep in mind that we'll need to upgrade nodegit more later to pick up libgit2 changes.  But it's easier to do smaller upgrades, so I thought I would start with this work.


novalis:
Closed because it's included in #533

PR #528, opened 2018-04-05
Rework rebase

bpeabody:
Addresses https://github.com/twosigma/git-meta/projects/2#card-8117622

Note: I still need to do manual/integration testing of this, but I think it's ready to start reviewing otherwise.

PR #533, opened 2018-04-10
git meta rm

novalis:
This is on top of #528, because #529 doesn't work without it.


bpeabody:
So this depends on #529 too?

novalis:
Well, it includes the work done in #529.

node/lib/util/rm.js
```diff
@@ -0,0 +1,482 @@
...
+    const noSuchFile = function() {
+        throw new UserError(`pathspec '${pathToRemove}' did not match any \
+files`);
+    };
+
+    if (pathToRemove.endsWith("/")) {
+        pathToRemove = pathToRemove.substring(0, pathToRemove.length - 1);
+    }
+
+    const idx = binarySearch(index.entries(), pathToRemove,
```
bpeabody: I don't see anything in the documentation to indicate that the array returned by `entries` is sorted. Am I missing something? If not, would `getByPath` work?

node/lib/util/rm.js
```diff
@@ -0,0 +1,482 @@
...
+    assert.isBoolean(force);
+
+    //note that "submodule" here is something of a misnomer in case 4
+    const bySubmodule = new Map();
+    const index = yield repo.index();
+    for (const path of paths) {
+        const spec = yield exports.getRemovalSpec(repo, index, path,
+                                                  recursive);
+        const submodule = spec.pathFromRoot;
+        if (submodule in bySubmodule) {
```
bpeabody: I'm a little surprised that `in` works; I think you'd want to use the `has` method with `Map`.

node/lib/util/rm.js
```diff
@@ -0,0 +1,482 @@
...
+            bySubmodule.get(submodule).push(spec);
+        }
+        else {
+            bySubmodule.set(submodule, [spec]);
+        }
+    }
+
+    for (const specs of bySubmodule.values()) {
```
bpeabody: Getting into the spirit of these new constructs, slightly cleaner (and more efficient, though it doesn't really matter): `for (const [_, specs] of bySubmodules) {`

node/lib/util/rm.js
```diff
@@ -0,0 +1,482 @@
...
+    });
+    const metaStaged = new Set(Object.keys(metaRepoStatus.staged));
+    const metaWorkdir = new Set(Object.keys(metaRepoStatus.workdir));
+
+    const badFiles = [];
+    const removedSubmodules = [];
+    const removedFiles = [];
+    const rootLevelSpecs = [];
+    const submoduleChanges = [];
+    for (const specs of bySubmodule.values()) {
```
bpeabody: and here too, `for (const [_, specs] of bySubmodule) {`

node/test/util/rm.js
```diff
@@ -0,0 +1,396 @@
...
+    });
+
+    describe("rmPaths", function () {
+        const cases = {
+            // Cases 3 and 4
+            "rm everything from empty repo": {
+                initial: "x=S:I README.md",
+                paths: [""],
+                fails: true,
+                expect: "x=S:I README.md",
```
bpeabody: If you pass `undefined` for expected, the implication is that you expect the repos to be unchanged, so you could leave this out.

node/test/util/rm.js
```diff
@@ -0,0 +1,396 @@
...
+            },
+            "rm a clean submodule by name": {
```
bpeabody: I don't see a submodule in this test case.

node/test/util/rm.js
```diff
@@ -0,0 +1,396 @@
...
+                initial: "x=S:C2-1 x/y/z=foo;W x/q/z=bar;Bmaster=2",
+                paths: ["x/y/z"],
+                expect: "x=S:C2-1 x/y/z=foo;W x/q/z=bar;I x/y/z;Bmaster=2",
```
bpeabody: Another simplification: the repo type "E" (for "expected") starts out as the case used initially, so you could simplify "expect" to: `x=E:I x/y/z` -- just specifying what's changed.

node/test/util/rm.js
```diff
@@ -0,0 +1,396 @@
...
+            },
+            "rm an open submodule": {
+                initial: "sub=S:C8-1 x/y/a=foo,x/y/z=foo;Bmaster=8|" +
+                    "x=S:C2-1 d=Ssub:8;" +
+                    "Bmaster=2;Od",
+                paths: ["d"],
+                expect: "sub=S:C8-1 x/y/a=foo,x/y/z=foo;Bmaster=8|" +
+                    "x=S:C2-1 d=Ssub:8;" +
+                    "Bmaster=2;I d"
+            },
```
bpeabody: Similarly, if you leave a repo out from the expected case, it's assumed to be unchanged, so this `expect` case can be simplified to `x=I d`.

node/test/util/rm.js
```diff
@@ -0,0 +1,396 @@
...
+                    isRootLevel: false,
+                    pathFromRoot: "d",
+                    pathInSubmodule: "x/y/z"
+                }
+            },
+            "rm a nonexistent dir from an open submodule": {
+                initial: "sub=S:C2-1 x/y/a=foo,x/y/z=foo;Bmaster=2|" +
+                    "x=S:C3-1 d=Ssub:2;" +
+                    "Bmaster=3;Od",
+                path: "d/q",
```
bpeabody: The parser is more graceful about line breaks now, so you can write this using the "`" string syntax that allows for embedded eols rather than string concatenation.

node/lib/util/rm.js
bpeabody: (nvm the `getByPath` suggestion)

novalis: The index is stored on disk sorted, and in regular git code, it is standard to assume that the in-memory representation is as well.  I don't see a reason why either libgit2 or nodegit would change the ordering, so I think this is probably safe despite the lack of doco.

node/lib/util/rm.js
novalis: I think values is lazy.

node/lib/util/rm.js
bpeabody: Makes sense.

node/lib/util/rm.js
bpeabody: Ah, OK. Well, I guess it doesn't matter much either way.

node/lib/util/rm.js
bpeabody: ... though you might want to put a comment saying as much there

node/lib/util/rm.js
```diff
@@ -0,0 +1,482 @@
...
+ * @return {NodeGit.Repo}  return.submoduleRepo
+ */
+exports.getRemovalSpec = co.wrap(function *(repo, index, pathToRemove,
+                                             recursive) {
+    assert.instanceOf(repo, NodeGit.Repository);
+    assert.instanceOf(index, NodeGit.Index);
+    assert.isString(pathToRemove);
+    assert.isBoolean(recursive);
+
+    const noSuchFile = function() {
```
bpeabody: `function noSuchFile()`

node/lib/util/rm.js
```diff
@@ -0,0 +1,482 @@
...
+ * @param {Boolean} recursive If false, and the path to remove is a directory,
+ * give the "can't remove a dir without -r" message
+ *
+ * @return {Object}
+ * @return {Boolean} return.isRootLevel -- distinguishes cases 1/2 from 3/4
+ * @return {String}  return.pathFromRoot to the submodule
+ * @return {String}  return.pathInSubmodule
+ * @return {NodeGit.Repo}  return.submoduleRepo
+ */
+exports.getRemovalSpec = co.wrap(function *(repo, index, pathToRemove,
```
bpeabody: As this started to soak in today, I found it very unfortunate that this was so much more complicated than `add`, which needs to do essentially the same logic and is only 77 lines of code (a little more than 1/10th of this). Of course, `add` has the advantage that it is looking for only *changed* files, and so can take utilize the `pathspec` in, e.g., `NodeGit.Diff.treeToWorkdir`.

However, I think with a little tweaking you could do the same thing as `add`. It might require some refactoring of `StatusUtil.getRepoStatus` and/or `DiffUtil.getRepoStatus`, depending on how you want to slice it, but at the crux of it you'd be operating based on a `Diff` constructed by running `NodeGit.Diff.treeToIndex` against an *empty* tree. The big advantage here is that you can let libgit2 do the nasty work of handling the `pathspec`. You could probably save a lot of effort even if you used `NodeGit.Diff.treeToIndex` directly rather than attempting to factor `DiffUtil` or `StatusUtil`.

Up to you, but I'll let you think on it before giving more feedback.

node/lib/util/rm.js
```diff
@@ -0,0 +1,416 @@
...
+        //a file that doesn't exist is considered OK for rm --cached, too
+        return null;
+    }
+
+    if (entry.mode === FILEMODE.COMMIT) {
+        // this is a submodule so we don't check the working tree
+        return new Problem(displayPath, "staged", "submodule");
+    }
+
+    // FIXME: this wastes memory for large objects
```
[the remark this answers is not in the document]
novalis: That has the side effect of adding the file to the object database.  Is that OK with you?

node/lib/util/rm.js
```diff
@@ -0,0 +1,416 @@
...
+                assert.equal("ENOENT", e.code);
+            }
+            if (stat !== null) {
+                if (stat.isDirectory()) {
+                    yield fs.rmdir(fullpath);
+                } else {
+                    yield fs.unlink(fullpath);
+                }
+            }
+            try {
```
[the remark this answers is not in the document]
novalis: good point

novalis:
I rebased and addressed the things other than TreeUtil.hashFile.

node/lib/util/rm.js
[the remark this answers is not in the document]
bpeabody: Yes, I'm OK with it.  Maybe it's not ideal, but my guess is that git is going to use a pretty efficient algorithm to compute the SHAs (that doesn't require loading the entire file into memory), and in any case, the SHA of the file in the index has already been computed.

bpeabody:
(I'm not strongly opposed to your original solution. The suggestion to use `hashFile` was just in answer to your concern about performance)

node/lib/util/rm.js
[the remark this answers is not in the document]
bpeabody: Yet another possibility would be `NodeGit.Diff.indexToWorkdir`, using the path in question as the only path in the `pathspec`.

bpeabody:
@novalis looks like we're failing the linter. I'll review otherwise.

node/lib/cmd/rm.js
```diff
@@ -0,0 +1,110 @@
+/*
```
bpeabody: Looks like this got folded into the wrong commit.

novalis:
fixed rebase and lint

Issue #517, opened 2018-03-23
flaky test continuing rebase


bpeabody:
Seen in unrelated change: https://github.com/twosigma/git-meta/pull/515#pullrequestreview-106602311
```
  1) rebase continueSubmodules rebase in a sub, was conflicted:
     Error: for repo x
In index
for path s expected Submodule(url=a, sha=qs) but got Submodule(url=a, sha=r)
for open submodule s
missing commit qs
HEAD is r but expected qs
In index
unexpected change to q: q
      at Object.exports.assertEqualRepoMaps (lib/util/repo_ast_util.js:644:15)
      at Object.<anonymous> (lib/util/repo_ast_test_util.js:352:17)
      at next (native)
      at onFulfilled (node_modules/co/index.js:65:19)
```

PR #541, opened 2018-04-16
Sparse reset

bpeabody:
Stop using `NodeGit.reset` to reset the meta-repo.


node/lib/util/git_util.js
```diff
@@ -942,6 +942,7 @@ exports.getUrlFromRemoteName = co.wrap(function *(repo, remoteName) {
 });
 
 /**
+<<<<<<< HEAD
```
novalis: That doesn't look good.

node/lib/util/submodule_config_util.js
```diff
@@ -70,6 +78,106 @@ const CONFIG_FILE_NAME = "config";
...
+    const newConfigLines = [];
+    const searchString = "[submodule \"" + submoduleName + "\"]";
+
+    let found = false;
+    let inSubmoduleConfig = false;
+
+    // Loop through the file and push, onto 'newConfigLines' any lines that
+    // aren't part of the bad submodule section.
+
+    for (let i = 0; i < configLines.length; ++i) {
```
novalis: for const line of configLines?

node/lib/util/submodule_config_util.js
```diff
@@ -70,6 +78,106 @@ const CONFIG_FILE_NAME = "config";
...
+
+    // If we didn't find the submodule, don't write the data back out.
+
+    if (found) {
+        newConfigLines.push("");  // one more new line
+        fs.writeFileSync(configPath, newConfigLines.join("\n"));
+    }
+
+    // Silence warning about no yield statement.
+    yield (Promise.resolve());
```
novalis: why not just use the async fs-promise funcs to read the file?

node/test/util/close_util.js
```diff
@@ -39,8 +39,8 @@ const RepoASTTestUtil = require("../../lib/util/repo_ast_test_util");
 describe("close_util", function () {
     describe("close", function () {
         // Don't need to test that deinitialization works, that's tested in
-        // deinit_util; just need to see that we handle paths and dirty
-        // sumobulde situations.
+        // submodule_config_util; just need to see that we handle paths and
+        // dirty sumobulde situations.
```
novalis: spelling of submodule

node/lib/util/reset.js
```diff
@@ -69,49 +75,69 @@ function getType(type) {
...
+            }
         }
-        const repo = yield opener.getSubrepo(name);
-        const status = yield StatusUtil.getRepoStatus(repo, {
-            showMetaChanges: true,
+    }
+
+    // Tidy up directories when not in sparse mode.
+
+    if (!(yield SparseCheckoutUtil.inSparseMode(repo))) {
```
novalis: If we're in sparse mode, why don't we do this?

node/lib/util/git_util.js
bpeabody: how embarrassing!

node/lib/util/submodule_config_util.js
bpeabody: thanks for the reminder

node/lib/util/submodule_config_util.js
bpeabody: I use the synchronous methods when mutating the config file to avoid race conditions.

bpeabody: (will add a note explaining this)

node/lib/util/reset.js
bpeabody: Well spell out reason in comment.

bpeabody:
Think I hit everything

novalis:
Are those test-failures expected?

bpeabody:
The test failures are not expected except for the continue one, for which I entered https://github.com/twosigma/git-meta/issues/517 -- I think there is a problem with libgit2 in rendering conflicts but need to do more investigation. I'll dig into the others.

bpeabody:
Nope, that's a different error on the conflict too!

bpeabody:
@novalis -- I cleared up all the tests, one of which did indicate a problem; I was not resetting submodules that were open (potentially with WD/Index changes) but unchanged HEAD. Also, in digging into that I I determined that `ResetUtil.abort` was doing unnecessary work.

novalis:
Conveniently, these test fixes also seem to fix the rm tests.  I'm pushing that rebased on top of this.

bpeabody:
I'm curious what test failures you were seeing with `rm` that this cleared up, @novalis .

bpeabody:
Both are fixed now?

I have seen occasional segfaults too, typically only when I run lots of tests at once. If you can figure out why, that would be awesome.

novalis:
I try to always run the full suite before pushing.  Before these changes, running it would always segfault. After, I got two runs without segfaults (and none with).  And the other one is fixed too.

bpeabody:
Only thing I can guess is that reduced reliance on the `cacheSubmodules` "hack" has helped.

PR #544, opened 2018-04-18
Fix bug leaving dangling `tree_entry` objects.

bpeabody:
Causing a crash on write.  This is caused by a bug in nodegit, as
described in the comment for this change.


novalis:
Looks like there are two other uses of treebuilder in the code too.

bpeabody:
Let me fix up those other instances and make the nodegit ticket then.

bpeabody:
@novalis -- I linked to the nodegit ticket (there was already an open ticket for this). I see only one other instance where `NodeGit.Treebuilder` is referenced in git-meta: `test/util/read_repo_ast_util.js`; it didn't suffer from this problem.

PR #547, opened 2018-04-20
Test framework support for sparse checkout mode.

bpeabody:
We're not trying to support general sparse checkout, just our specific
mode, and we don't really care about non-submodule changes.  Overview:

1. Added `sparse` to RepoAST and RepoASTUtil
2. Support for reading and writing repos with in sparse mode.
3. Shorthand support -- '%' prefixed to a repos base type indicates
   sparse mode.

Addresses: https://github.com/twosigma/git-meta/projects/5#card-9115833


novalis:
Maybe we should have a documentation file about sparse?

PR #553, opened 2018-04-22
[issue#548] Print output even if hook returns non-zero status, give h…

### LINKED ISSUE #548: git meta swallows hook output if return code is non-zero.
Seems like this call returns reject promise:

https://github.com/twosigma/git-meta/blob/master/node/lib/util/hook.js#L56

So we prob should output it in the catch clause. While we here, lets also check if hook is executable.

I will look at it tonight.

zokir:
fixes #548 

also added a message if git hook is not executable, although actual git just ignores in this case.


bpeabody:
Failing test is a flake

bpeabody:
(unrelated to this PR)

PR #557, opened 2018-04-23
#556: git meta commit-shadow should allow a timestamp to be specified

s-xiang:


node/lib/cmd/commit-shadow.js
```diff
@@ -75,10 +75,16 @@ exports.configureParser = function (parser) {
         required: true,
         help: "commit message for shadow commits",
     });
+    parser.addArgument(["-t", "--timestamp"], {
+        type: Number,
+        required: false,
+        defaultValue: null,
+        help: "timestamp for shadow commits",
```
bpeabody: In what format?

node/lib/util/stash_util.js
```diff
@@ -536,13 +542,18 @@ const makeShadowCommitForRepo = co.wrap(function *(repo,
...
                                               message,
+                                              timestamp,
                                               includeMeta,
                                               includeUntracked) {
     assert.instanceOf(repo, NodeGit.Repository);
     assert.isString(message);
     assert.isBoolean(includeMeta);
     assert.isBoolean(includeUntracked);
 
+    if (null !== timestamp && Number.isNaN(timestamp)) {
```
bpeabody: This is a contract violation, not a user error. Try:

```JavaScript
if (null !== timestamp) {
    assert.isNumber(timestamp);
}
```

node/test/util/stash_util.js
```diff
@@ -820,4 +822,50 @@ x=E:Cm-1 s=Sa:s;Bm=m;Os Cs foo=bar!Bs=s!W foo=bar`
             }));
         });
     });
+    describe("shadowCommit", function () {
```
bpeabody: Extend the existing `"shadow"` test cases to support timestamp. You don't need to validate that two commits compare equal in the test driver, just that the one commit generated has the expected timestamp, if provided (you should validate this for sanity on the CL that the SHAs are the same, in case there's something weird libgit2 is doing).

node/lib/util/stash_util.js
s-xiang: Tried this way, in this way, it did not throw out an error. That is why I did like this.

node/lib/cmd/commit-shadow.js
s-xiang: Do we care about this format? if the user uses the same format, it should generate the same sha.

node/lib/util/stash_util.js
```diff
@@ -476,17 +476,20 @@ exports.list = co.wrap(function *(repo) {
  * @param {NodeGit.Repository} repo
  * @param {RepoStatus}         status
  * @param {String}             message
+ * @param {Number}             timestamp
```
bpeabody: call this `useEpochTimestamp`

node/lib/util/stash_util.js
```diff
@@ -536,12 +543,14 @@ const makeShadowCommitForRepo = co.wrap(function *(repo,
  */
 exports.makeShadowCommit = co.wrap(function *(repo,
                                               message,
+                                              includeTimestamp,
```
bpeabody: Ditto here, I think `useEpochTimestamp` makes more sense.

node/test/util/stash_util.js
```diff
@@ -785,9 +786,12 @@ x=E:Cm-1 s=Sa:s;Bm=m;Os Cs foo=bar!Bs=s!W foo=bar`
                 const meta = c.includeMeta;
                 const includeUntracked =
                         undefined === c.includeUntracked || c.includeUntracked;
+                const includeTimestamp =
+                        undefined === c.includeTimeStamp || c.includeTimeStamp;
                 const result = yield StashUtil.makeShadowCommit(
                                                              repo,
                                                              message,
+                                                             includeTimestamp,
```
bpeabody: Validate that the timestamp of the resulting commit is the epoch if useEpochTimestamp is used.

Issue #562, opened 2018-04-28
`TreeUtil` methods break when a submodule is changed into a tree (containing submodules)


bpeabody:
`buildDirectoryTree` needs to change to accept a `null` value as a parent and `writeTree` needs to make a new tree when it sees an entry that must be a parent but which is not actually a tree.

PR #563, opened 2018-04-28
Address path mode changes in `TreeUtil`

bpeabody:
Specifically, those where a "tree" (a.k.a. directory) changes into
something else and vice-versa.

Addresses: https://github.com/twosigma/git-meta/issues/562


bpeabody:
Added a fix to `StashUtil` for breakage introduced in nodegit upgrade that I noticed in test failure for this PR.

PR #564, opened 2018-04-29
Added synthic-ref-check whitelist.

lally:
This is neutral to the tests.  How do I add a test that alters config?  So I can validate that some URIs pass through the whitelist.

### DIFF
--- a/node/lib/util/synthetic_branch_util.js
+++ b/node/lib/util/synthetic_branch_util.js
@@ -54,6 +54,17 @@ function identity(v) {
     return v;
 }
 
+function SyntheticBranchConfig(whitelistPattern) {
+    if (whitelistPattern.length > 0) {
+        const whitelistRE = new RegExp(whitelistPattern);
+        this.whitelistTest = function(url) {
+            return whitelistRE.test(url);
+        };
+    } else {
+        this.whitelistTest = function() { return false; };
+    }
+}
+
     /**
      * (This has to be public so we can mock it for testing)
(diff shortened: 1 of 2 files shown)


node/lib/util/synthetic_branch_util.js
```diff
@@ -96,6 +96,29 @@ exports.urlToLocalPath = function *(repo, url) {
     }
 };
 
+/**
+ * Check that a given URL i son the URLs synthetic-ref-check whitelist, if
+ * such a whitelist exists.
+ * @async
+ * @param {NodeGit.Repository} repo The meta repository
+ * @param {String}             url The configured URL of hte submodule
```
novalis: spelling of 'the'

node/lib/util/synthetic_branch_util.js
```diff
@@ -96,6 +96,29 @@ exports.urlToLocalPath = function *(repo, url) {
...
+ * in the meta tree.
+ */
+function* skipCheckForURL(repo, url) {
+    assert.instanceOf(repo, NodeGit.Repository);
+    assert.isString(url);
+    const config = yield repo.config();
```
novalis: I am worried about doing this on every submodule url because I'm not sure it's cached -- could we extract it?

novalis:
I think you can just get the repo object in the test and directly alter the config using the usual nodegit functions

bpeabody:
I think I fixed the test failure you see here in https://github.com/twosigma/git-meta/pull/563

PR #576, opened 2018-05-02
initial stab at better push

novalis:
WIP -- it needs some comments and tests.  But let me know what you think.

### DIFF
--- a/node/lib/util/push.js
+++ b/node/lib/util/push.js
@@ -46,6 +46,120 @@ const SubmoduleConfigUtil = require("./submodule_config_util");
 const SyntheticBranchUtil = require("./synthetic_branch_util");
 const UserError           = require("./user_error");
 
+
+/**
+ * For a given proposed push, return a map from submodule to sha,
+ * excluding any submodules that the server likely already has.
+ *
+ * Start with the set of submodules that are either open, or that
+ * exist in .git/modules. Populate a pushMap from submodule name to
+ * commit (SHA) registered in the meta-commit being pushed.
+ *
+ * If the push target exists in remotes/$remote/$branch, remove from
+ * pushMap, any entry where the commit for a submodule already exists
+ * in the target.
+ *
+ * Note that for "exists" we mean the target points to exactly the
+ * commit needed or a descendant of that commit.
+ *
+ * If the user is pushing a branch, B, further reduce the size of
+ * pushMap by removing commits referenced in the branches that B pulls
+ * from and/or pushes to by default, if they exist and are different
+ * from the target branch.
+ *
+ * @async
+ * @param {NodeGit.Repository} repo
+ * @param {String}             remoteName
+ * @param {String}             source
+ * @param {String}             target
+ * @param {NodeGit.Commit}     commit
+ */
+exports.getPushMap = co.wrap(function*(repo, remoteName, source, target,
+                                       commit) {
+    assert.instanceOf(repo, NodeGit.Repository);
+    assert.isString(remoteName);
+    assert.isString(source);
+    assert.isString(target);
+    assert.instanceOf(commit, NodeGit.Commit);
+
+    const submoduleSet = new Set(yield SubmoduleUtil.listOpenSubmodules(repo));
+    for (const s of yield SubmoduleUtil.listAbsorbedSubmodules(repo)) {
+        submoduleSet.add(s);
+    }
+    const submodules = [...submoduleSet];
+    const pushMap = yield SubmoduleUtil.getSubmoduleShasForCommit(repo,
+                                                                  submodules,
+                                                                  commit);
+    const bareRepos = {};
+    for (const sub of Object.keys(pushMap)) {
+        const subBareRepo = yield SubmoduleUtil.getBareRepo(repo, sub);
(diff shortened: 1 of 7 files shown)


node/lib/util/push.js
```diff
@@ -84,23 +84,75 @@ exports.push = co.wrap(function *(repo, remoteName, source, target, force) {
     let remoteUrl = yield GitUtil.getUrlFromRemoteName(repo, remoteName);
 
     // First, push the submodules.
+    const annotatedCommit = yield GitUtil.resolveCommitish(repo, source);
```
bpeabody: I'd like to see the logic for listing what to push factored out into a separate (and separately-tested) method.

node/lib/util/push.js
```diff
@@ -84,23 +84,75 @@ exports.push = co.wrap(function *(repo, remoteName, source, target, force) {
...
+    const sha = annotatedCommit.id();
+    const commit = yield repo.getCommit(sha);
+
+    const submoduleSet = new Set(yield SubmoduleUtil.listOpenSubmodules(repo));
```
bpeabody: Is there a way you could construct this `Set` from the contents of two arrays without using a loop?

node/lib/util/push.js
```diff
@@ -84,23 +84,75 @@ exports.push = co.wrap(function *(repo, remoteName, source, target, force) {
...
+                `refs/remotes/${remoteName}/${tracking.remoteName}`);
+        }
+        if (tracking.pushRemoteName !== null) {
+            trackingBranches.add(
+                `refs/remotes/${remoteName}/${tracking.pushRemoteName}`);
+        }
+    }
+
+    for (const branch of trackingBranches) {
+        const annotated = yield GitUtil.resolveCommitish(repo, branch);
```
bpeabody: I think `resolveCommitish` is too heavy a tool for this; you have the exact ref name. I'd use `NodeGit.Reference.lookup`.

node/lib/util/push.js
```diff
@@ -84,23 +84,75 @@ exports.push = co.wrap(function *(repo, remoteName, source, target, force) {
...
+                    entry = yield tree.entryByPath(sub);
+                } catch (e) {
+                    //not found, OK
+                    continue;
+                }
+                // easy check: does the tracking branch's id equal the one
+                // in the pushMap?  If not, it would be possible that the
+                // tracking branch were *ahead*, but this is so unlikely as
+                // to be not worth opening up the repo to check
+                if (entry.id() === pushMap[sub]) {
```
bpeabody: I think you should open up the repo and check in the case that they're not equal. It would be great if you can come up with a way to do this without fully opening repos that are just absorbed -- I'll reuse this logic in places like `merge` where I sometimes have to fully open a repo just to decide that something could have been fast-forwarded.

It seems fairly likely that the SHAs from the upstream branch may have progressed before you've pushed to your remote.

node/lib/util/submodule_util.js
```diff
@@ -201,6 +202,62 @@ exports.getCurrentSubmoduleShas = function (index, submoduleNames) {
...
 
+const gitReservedNames = new Set(["HEAD", "FETCH_HEAD", "ORIG_HEAD",
+                                  "COMMIT_EDITMSG", "index", "config",
+                                  "logs", "rr-cache", "hooks", "info",
+                                  "objects", "refs"]);
+/**
+ * Return a list of submodules from .git/modules -- that is,
+ * approximately, those which we have ever opened.
+ */
+exports.listAbsorbedSubmodules = co.wrap(function*(repo) {
```
bpeabody: Will definitely want to see a test for this.

node/lib/util/push.js
bpeabody: Not necessarily; it seems as though there should be a better way but I can't think of one either.

novalis:
OK, updated the PR.

node/lib/util/push.js
```diff
@@ -46,6 +46,99 @@ const SubmoduleConfigUtil = require("./submodule_config_util");
...
+        try {
+            reference = yield NodeGit.Reference.lookup(repo, branch);
+        } catch (e) {
+            // if this ref doesn't exist, OK
+            continue;
+        }
+        const id = reference.target();
+        const trackingCommit = yield NodeGit.Commit.lookup(repo,
+                                                           id);
+        const tree = yield trackingCommit.getTree();
```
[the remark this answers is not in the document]
novalis: ok, one more version

node/lib/util/push.js
```diff
@@ -46,6 +46,113 @@ const SubmoduleConfigUtil = require("./submodule_config_util");
...
+            const commitToPush = yield subRepo.getCommit(pushMap[sub]);
+            let trackingCommit;
+            try {
+                trackingCommit = yield subRepo.getCommit(sha);
+            } catch (e) {
+                // We haven't fetched this commit in this submodule,
+                // so we can't do an ancestry check
+                continue;
+            }
+            if (NodeGit.Graph.descendantOf(subRepo, commitToPush,
```
bpeabody: Do you need to check if it's descendantOf or equal to? If I'm wrong, ignore me, but if I'm right, there should probably be a test case that fails because this is missing.

PR #586, opened 2018-05-09
use increment instead of epoch for shadow times

bpeabody:
Commits with old times confuse (as in makes very expensive) Git's fetch
logic, so instead of using an epoch time to make sure subsequent shadow
commits of unchanged code have the same sha, we'll use the time of HEAD
+ 1.


s-xiang:
LGTM!

PR #600, opened 2018-05-21
sync sparse-checkout & index after open/close

bpeabody:
The current strategy for sparse mode was to exclude in the
sparse-checkout file and index all files besides `.gitmodules`.  This
strategy was fine as far as `git-meta` was concerned generally, but it
confused Git; since `git-meta` defers to Git for many commands (e.g.,
`diff`), and since we generally want to allow people to use normal Git
commands and maintain our repo in a recognizable state, the current
strategy is not sufficient.

Instead, we need to maintain the invarian that "open" submodules are
seen as visible in the sparse mode (not skipped) and "closed" submodules
are seen as invisible (skipped).

Severl changes were needed to make this happen:

* We now consider a repo to be "inSparseMode" if the config is set and
  don't also check to see if `".gitmodules\n"` is the contents of the
  `sparse-checkout` file.  It was a little pedantic to add that to the
  check to begin with, and now we're going to have an arbitrary set of
  paths.
* When opening a submodule, we add its path to `sparse-checkout`.
* `SubmoduleConfigUtil.deinit` now removes submodules from
  `sparse-checkout`.  Since this would be O(N^2) (e.g. if you're closing
  all submodules) and in all cases we generally know up front the list
  of subs to remove, `deinit` now takes a list of subs to deinitialize.
* Opening and closing subs now requires an update to the index, since
  libgit2 (and I'm guessing not even Git) will do what we need by
  default.  To facilitate this, I updated the contracts of the
  open/close methods, and moved `writeMetaIndex` down into
  `SparseCheckoutUtil`.
* Call `writeMetaIndex` from stash_util, which didn't need to update the
  index previously
* And also from `CloseUtil`.

### DIFF
--- a/node/lib/util/sparse_checkout_util.js
+++ b/node/lib/util/sparse_checkout_util.js
@@ -57,9 +57,7 @@ exports.getSparseCheckoutPath = function (repo) {
 
 /**
  * Return true if the specified `repo` is in sparse mode and false otherwise.
- * A repo is in sparse mode iff: `core.sparsecheckout` is true and the contents
- * of `.git/info/sparse-checkout` is exactly ".gitmodules\n".  We can do
- * something more general purpose later if we deem it useful.
+ * A repo is in sparse mode iff: `core.sparsecheckout` is true.
  *
  * @param {NodeGit.Repository} repo
  * @return {Bool}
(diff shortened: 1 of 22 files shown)


node/lib/util/sparse_checkout_util.js
```diff
@@ -67,23 +67,16 @@ exports.getSparseCheckoutPath = function (repo) {
...
-    }
-    let content;
-    try {
-        content = yield fs.readFile(exports.getSparseCheckoutPath(repo),
-                                    "utf8");
-    } catch (e) {
-        return false;                                                 // RETURN
-    }
-    return content === ".gitmodules\n";
+    return (yield ConfigUtil.configIsTrue(repo, "core.sparsecheckout")) ||
```
novalis: Does configIsTrue not already return a boolean?

node/lib/util/sparse_checkout_util.js
```diff
@@ -94,3 +87,97 @@ exports.setSparseMode = co.wrap(function *(repo) {
...
+    // libgit2 will not.
+
+    if (yield exports.inSparseMode(repo)) {
+        const sparseCheckout = exports.readSparseCheckout(repo);
+        const sparseSet = new Set(sparseCheckout.split("\n"));
+        const NORMAL = 0;
+        for (const e of index.entries()) {
+            if (NORMAL === NodeGit.Index.entryStage(e)) {
+                if (sparseSet.has(e.path)) {
+                    e.flagsExtended ^= SKIP_WORKTREE;
```
novalis: We're assuming here that it's presently set and we want it unset -- but what if it wasn't previously set?  You want e.flagsExtended &= ~SKIP_WORKTREE, I think.

node/lib/util/sparse_checkout_util.js
bpeabody: If the value is not configured it returns null.

node/lib/util/sparse_checkout_util.js
bpeabody: Yeah, you're right.

bpeabody:
Updated

node/lib/util/sparse_checkout_util.js
novalis: That function, then, is probably misnamed, esp since it will usually be used in a boolean context.  But no need to fix it in this PR I guess.

bpeabody: I did update the comment, which was still referencing the previous behavior.

PR #603, opened 2018-05-21
Amend merge

novalis:


node/lib/util/commit.js
```diff
@@ -922,14 +945,20 @@ exports.getAmendStatus = co.wrap(function *(repo, options) {
         const change = changes[name];
         let currentSub = submodules[name];
         let old = null;
-        if (undefined !== change) {
+        const isMerged = Object.keys(merged).includes(name);
```
bpeabody: `name in merged` would be much more efficient (should be a hash lookup)

node/lib/util/commit.js
```diff
@@ -889,24 +889,47 @@ exports.getAmendStatus = co.wrap(function *(repo, options) {
...
-               yield SubmoduleConfigUtil.getSubmodulesFromCommit(repo, parent);
-    }
-    const diff =
-               yield NodeGit.Diff.treeToTree(repo, parentTree, headTree, null);
-    const changes = yield SubmoduleUtil.getSubmoduleChangesFromDiff(diff,
-                                                                    true);
+    const oldUrls = {};
+    const parents = yield head.getParents();
+
+    const merged = {};
```
bpeabody: Need more documentation about what your algorithm is here -- e.g., what are `merged` and `changes` for? I think it might be a good point to factor this logic into a new method, e.g., `computeAmendSubmoduleChanges`.

bpeabody: I'd also recommend that instead of having a `merged` and `changes` approach like this, you return a single recording the correct new and old SHAs so you don't have a special case for merge later.

novalis: I think I need a special case for merge.  For submodules changed in this commit, we do want to amend them. For ones not changed, we don't.

bpeabody: But I think the algorithm for that is to amend submodule commits whose SHA doesn't exist in the tree of any immediate parent. The case of having one parent doesn't seem special.

PR #606, opened 2018-05-30
Fix push in a sparse repo

novalis:
This works around a bug in git where push attempts to cd into a worktree
even though itn fact it could work juts fine in a bare repo.


novalis:
(Testing: tested manually inside TS and I believe it works)

PR #610, opened 2018-06-07
Checkout paths

### LINKED ISSUE #368: support <path> parameter for checkout

novalis:
Fixes #368

### DIFF
--- a/node/lib/util/checkout.js
+++ b/node/lib/util/checkout.js
@@ -44,6 +44,7 @@ const Reset              = require("./reset");
 const RepoStatus         = require("./repo_status");
 const SparseCheckoutUtil = require("./sparse_checkout_util");
 const StatusUtil         = require("./status_util");
+const StashUtil          = require("./stash_util");
 const SubmoduleFetcher   = require("./submodule_fetcher");
 const SubmoduleUtil      = require("./submodule_util");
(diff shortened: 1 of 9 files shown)


node/lib/util/checkout.js
```diff
@@ -382,18 +403,27 @@ A branch named ${colors.red(name)} already exists.`);
...
+                // If we didn't resolve anything from `committish`, try it
+                // as a file.
+                files.splice(0, 0, committish);
+                // We would now like to resolve from the index, but
+                // libgit2 doesn't support the ":0" shorthand, and our
+                // remaining code all assumes an actual commit, so
+                // we're just going to make one.  The cache-tree ought to
+                // make this relatively cheap.
+
+                annotated = yield makeShadowCommitFromIndex(repo);
```
bpeabody: Clever! I'm not sure that this is less complex than allowing `checkoutFiles` to work on an index if `commit` is null, but I don't have a strong opinion.

When we want to support `--ours` and `--theirs`, can we make these options to `makeShadowCommitFromIndex`?

novalis: I considered something like null, or special-casing ":0" (and :1 &c), but decided this was simpler.  And I think --ours and --theirs will fall out naturally, just as you suggest.

PR #617, opened 2018-06-20
Work around libgit2 merging bug

novalis:


node/lib/util/cherry_pick_util.js
```diff
@@ -240,7 +273,20 @@ exports.computeChanges = co.wrap(function *(repo, index, targetCommit) {
...
         return new ConflictEntry(entry.mode, entry.id.tostrS());
     }
+
+    const mergeBases = yield GitUtil.mergeBases(repo, head, targetCommit);
+    const submoduleShasInBases = {};
+    for (const base of mergeBases) {
+        const commit = yield NodeGit.Commit.lookup(repo, base);
+        const names = yield SubmoduleUtil.getSubmoduleNamesForCommit(repo,
+                                                                     commit);
+        const shas = yield SubmoduleUtil.getSubmoduleShasForCommit(repo, names,
```
bpeabody: This is actually a fairly expensive operation; it has to call into libgit2 once for each submodule; it can be ~1s to do this for a meta-repo with thousands of submodules and I've gone to lengths in recent changes to avoid doing this for all subs.

Could you tweak this solution so that we only need to read the SHAs for subs in add-add situations? We should avoid calling `mergeBases` unless we're in this situation as well as we've seen that it can be pretty expensive (e.g., 800ms) even for branches just a week apart.

node/lib/util/cherry_pick_util.js
```diff
@@ -177,6 +178,63 @@ exports.containsUrlChanges = co.wrap(function *(repo, commit, baseCommit) {
...
+    if (data.submoduleShasInBases === undefined) {
+        const head = data.head;
+        const targetCommit = data.targetCommit;
+        const mergeBases = yield GitUtil.mergeBases(repo, head, targetCommit);
+        const submoduleShasInBases = {};
+        for (const base of mergeBases) {
+            const commit = yield NodeGit.Commit.lookup(repo, base);
+            const names = yield SubmoduleUtil.getSubmoduleNamesForCommit(
+                repo, commit);
+            const shas = yield SubmoduleUtil.getSubmoduleShasForCommit(repo,
```
bpeabody: This is closer in that we populate the data lazily (only when encountering an add/add situation), but there's still no need to read the SHAs for all (thousands) of submodules, just the ones in an add/add scenario.

My recommendation: cache the result of `mergeBases`, but do the she-reading on-demand.

Issue #615, opened 2018-06-13
`add` clears conflict markers on all files in a submodule it touches


bpeabody:
Repro:
```bash
$ write-repos -o 'a=B:Cx-1 a=3,b=4;Ca-x a=9,b=8;Cb-x a=8,b=9;Bx=x;Ba=a;Bb=b|x=U:C3-2 s=Sa:x;C4-3 s=Sa:a;C5-3 s=Sa:b;Bmaster=4;Bfoo=5'
$ cd x
$ git meta merge foo
Opening s on ced42a31aa2b22b3c35fce24195432c1f2662d35.
Submodule s: merging commit 5113bcbfc4165d9f055a302cbfa6a06c1c15f76f.
Submodule s is conflicted.
$ git meta status
On branch master.
A merge is in progress.
  (after resolving conflicts mark the corrected paths
   with 'git meta add', then run "git meta merge --continue")
  (use "git meta merge --abort" to check out the original branch)
Changes to be committed:
  (use "git meta reset HEAD <file>..." to unstage)

        conflicted:   s/a
        conflicted:   s/b
$ git meta add s/a
$ git meta status
On branch master.
A merge is in progress.
  (after resolving conflicts mark the corrected paths
   with 'git meta add', then run "git meta merge --continue")
  (use "git meta merge --abort" to check out the original branch)
Changes to be committed:
  (use "git meta reset HEAD <file>..." to unstage)

        modified:     s/a
        modified:     s/b
```
Note that we added only `s/a` yet the conflict on `s/b` was also removed.

bpeabody:
After a little investigation, I can see that the problem is that `StatusUtil.getRepoStatus`, on which `add` relies, loads up all conflicts in the results, ignoring paths. We are doing this on line 481 where we call `exports.readConflicts`, to read in the conflict status of the submodules which would otherwise be ignored. I think we need to make this method honor the paths.

PR #621, opened 2018-06-27
fixed issue #615, add by path

zokir:
closes issue #615


novalis:
Just confirming: this test was red before and is now green?

zokir:
Yes, I made sure that test by Path was failing before applying this change.

PR #626, opened 2018-07-06
Added explicit module post-checkout hook execution

zokir:
To execute post-checkout hook

### DIFF
--- a/node/lib/util/git_util.js
+++ b/node/lib/util/git_util.js
@@ -710,6 +710,13 @@ exports.setHeadHard = co.wrap(function *(repo, commit) {
     yield NodeGit.Checkout.tree(repo, commit, {
         checkoutStrategy: NodeGit.Checkout.STRATEGY.FORCE,
     });
+    const execString = `\
+git -C '${repo.workdir()}' checkout ${commit.id()} -f `;
+    try {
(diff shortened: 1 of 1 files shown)


novalis:
That is: why not just run the hook manually?

bpeabody:
Yes, that is a good point too, @novalis . Further, I wonder if there are other operations we're performing to submodules besides this where hooks are not triggered. I suspect a more systematic approach is needed to do this "right". If this is just a hack to fix an internal situation for us, I'd rather patch it internally rather than doing something half-baked here, especially since that situation will be going away soon.

novalis:
I think it's generally the right thing to do.  Maybe we also would want to run more hooks elsewhere (as we now do for other operations), but I don't think it's urgent to do those now.  It's sufficient to do them as we notice them.  This is the big one.

bpeabody:
OK, I can live with that. Are we agreed that we will continue to call libgit2 checkout here, not shell out to git, but do invoke the hook if it exists?

I imagine there is some logic from invoking hooks in the meta-repo that we should be reusing?

bpeabody:
We don't need this any more after https://github.com/twosigma/git-meta/pull/627

PR #635, opened 2018-07-23
Fix calls to post-checkout hook.

novalis:
Make post-checkout hook run as "branch checkout" in detached head
state.  Make it run at all in file checkout mode.

Per the git docs, "branch checkout" (passing "1" to the post-checkout
hook) does not mean "branch, as opposed to detached".  It means
"branch, as opposed to files".


novalis:
I had thought there was another but, but I was wrong.  While I was in the code, tho, I noticed this.

s-xiang:
LGTM

PR #646, opened 2018-08-03
Add flag to Commit-Shadow that allows specifying the set of sub-repos…

blevz:
… to consider for the resulting shadow commit, add tests to stash_util.shadow

### DIFF
--- a/node/lib/cmd/commit-shadow.js
+++ b/node/lib/cmd/commit-shadow.js
@@ -50,12 +50,14 @@ exports.helpText =
  * @property {String}
  */
 exports.description  = `Create a "shadow" commit containing all local
-modifications (including untracked files if '--include-untracked' is specified)
-to all sub-repos and then print the SHA of the created commit.  If there are no
-local modifications, print the SHA of HEAD.  Do not modify the index or update
-HEAD to point to the created commit.  Note that this command ignores
-non-submodule changes to the meta-repo.  Note also that this command is meant
-for programmatic use and its output format is stable.`;
+modifications (including untracked files if '--include-untracked' is specified,
+include only files in the specified directories following '--include-subrepos' 
+if it is specified and non-empty) to all sub-repos and then print the SHA of 
+the created commit.  If there are no local modifications, print the SHA of 
+HEAD.  Do not modify the index or update HEAD to point to the created commit.  
+Note that this command ignores non-submodule changes to the meta-repo. 
+Note also that this command is meant for programmatic use and its output
+format is stable.`;
 
 /**
  * Configure the specified `parser` for the `commit` command.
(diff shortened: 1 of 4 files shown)


node/lib/cmd/commit-shadow.js
```diff
@@ -90,6 +90,12 @@ exports.configureParser = function (parser) {
         defaultValue: false,
         help: "use timestamp of HEAD + 1 instead of current time",
     });
+    parser.addArgument(["-s", "--include-subrepos"], {
+        type: "string",
+        required: false,
+        defaultValue: "",
+        help: "only include specified sub-repos, includes all if left empty",
```
bpeabody: No strong opinion on this but would like feedback from @novalis -- should empty-list really imply all? It seems to me that the natural behavior for this (because it would require no special cases or documentation) would be for the empty list to mean none.

node/lib/util/stash_util.js
```diff
@@ -571,6 +571,7 @@ const makeShadowCommitForRepo = co.wrap(function *(repo,
  * @param {Bool}               useEpochTimestamp
  * @param {Bool}               includeMeta
  * @param {Bool}               includeUntracked
+ * @param {Object}             includedSubrepos
```
bpeabody: Please extend the contract of `makeShadowCommit` to include this parameter.

node/lib/cmd/commit-shadow.js
novalis: Well, it sure makes migration easier!

But I take your point that it's a little out-of-sync with the rest of git, so probably it's worth the slight hassle.

bpeabody: I don't think it affects migration, does it? The default when this argument is omitted would be to include all submodules. I'm just talking about the behavior when you use this argument but provide no paths to it.

blevz: I agree, empty array == all is a little confusing but it matches the convention used in getRepoStatus and the alternatives (using null or undefined) seem more kludgey, especially because they would just be converted to an empty list to be passed to getRepoStatus.  +1 to getting some feedback from @novalis

bpeabody: Yes, I think it make sense for the `makeShadowCommitForRepo` method, but not necessarily for the CL API -- but I don't feel that strongly about it. If you and @novalis are in agreement that it's OK as is, it's fine with me.

blevz: We could have it error out or return the sha for head if the flag is set with no values passed

bpeabody: I think you can configure argparse to that it fails unless you have at least one arg. I like that solution.

PR #647, opened 2018-08-03
Added stderr and disabled buffering in hooks

zokir:
What do you guys think about this change?

Basically, I want to see the output of a hook in real time instead of waiting till hook execution finishes. 
This is important when hooks executes long running operations like build/tests etc. to let user know what is going on.
Additionally, I want to see the stderr output.

### DIFF
--- a/node/lib/util/hook.js
+++ b/node/lib/util/hook.js
@@ -31,30 +31,35 @@
 "use strict";
 
 const assert    = require("chai").assert;
-const ChildProcess = require("child-process-promise");
+const spawn     = require("child-process-promise").spawn;
 const co        = require("co");
 const path      = require("path");
 const process   = require("process");
+const fs        = require("fs");
 
 /**
  * Run git-meta hook with given hook name.
+ * Return 0 on success, non-zero status otherwise.
  * @async
  * @param {String} name
  * @param {String[]} args
- * @return {String}
+ * @return {int}
  */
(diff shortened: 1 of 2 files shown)


node/lib/util/hook.js
```diff
@@ -43,17 +44,26 @@ const process   = require("process");
...
+        return absPath + " hook does not exist.";
+    }
+
     try {
         process.chdir(repo.workdir());
-        const result = yield ChildProcess.execFile(absPath, args);
-        console.log(result.stdout);
+        const result = spawn(absPath, args);
+        result.stdout.on("data", function(data) {
+            console.log(data.toString());
```
novalis: Doesn't this end up inserting additional newlines?

novalis: And, actually, is there a way to just hook our stdout/stderr up to the process so that we don't have to read/write?

node/package.json
```diff
@@ -31,6 +31,7 @@
     "binary-search": "",
     "chai": "",
     "child-process-promise": "",
```
novalis: It's a shame to have two child-process libraries.  Is there any way we can consolidate to one?  If not, let's ship this.

bpeabody:
Yes, I’d like to get an answer to David’s question. Why do we need a new child process library?

zokir:
Not intentional. I must have been playing with spawnSync and then forgot to go back. Reverted.

PR #657, opened 2018-08-23
replace NodeGit.Repository.defaultSignature with our own function

novalis:
NodeGit's occasionally returns unknown@example.com, for unknown
reasons, and this is bad.


novalis:
We continue to use repo.defaultSignature() in the tests, and we compare against it, so this function is tested, but I can add a specific test later.

PR #660, opened 2018-09-11
Initial implementation of tool to stitch repos

bpeabody:
standalone script in `lib/stitch.js`

Also has the ability to join submodule histories, which I would like to be able to use.

### DIFF
--- a/node/lib/stitch.js
+++ b/node/lib/stitch.js
@@ -1,0 +1,148 @@
+#!/usr/bin/env node
+/*
+ * Copyright (c) 2017, Two Sigma Open Source
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of git-meta nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+const ArgumentParser = require("argparse").ArgumentParser;
+const co             = require("co");
+
+const StitchUtil = require("./util/stitch_util");
+
+const description = `Stitch together the specified meta-repo commitish in \
+the specified repo.`;
+
+const parser = new ArgumentParser({
+    addHelp: true,
+    description: description
+});
+
+parser.addArgument(["--no-fetch"], {
+    required: false,
+    action: "storeConst",
+    constant: true,
+    defaultValue: false,
+    help: `If provided, assume commits are present and do not fetch.`,
+});
+
+parser.addArgument(["-t", "--target-branch"], {
+    required: false,
+    type: "string",
+    defaultValue: "master",
+    help: "Branch to update with committed ref; default is 'master'.",
+});
+
+parser.addArgument(["-j"], {
+    required: false,
+    type: "int",
+    help: "number of parallel operations, default 8",
+    defaultValue: 8,
+});
+
+parser.addArgument(["-c", "--commitish"], {
+    type: "string",
+    help: "meta-repo commit to stitch, default is HEAD",
+    defaultValue: "HEAD",
+    required: false,
+});
+
+parser.addArgument(["-u", "--url"], {
+    type: "string",
+    defaultValue: null,
+    help: `location of the origin repository where submodules are rooted, \
+required unless '--no-fetch' is specified`,
+    required: false,
+});
+
+parser.addArgument(["-r", "--repo"], {
+    type: "string",
+    help: "location of the repo, default is \".\"",
+    defaultValue: ".",
+});
+
+parser.addArgument(["-k", "--keep-as-submodule"], {
+    type: "string",
+    help: `submodules whose paths are matched by this regex are not stitched, \
+but are instead kept as submodules.`,
+    required: false,
+    defaultValue: null,
+});
+
+parser.addArgument(["--skip-empty"], {
+    required: false,
+    action: "storeConst",
+    constant: true,
+    defaultValue: false,
+    help: "Skip a commit if its  tree would be the same as its first parent.",
+});
+
+parser.addArgument(["--root"], {
+    type: "string",
+    help: "When provided, run a *join* operation that creates a new history \
+joining those of the submodules under the specified 'root' path.  \
+In this history those paths will be relative to 'root', i.e., they will not \
+be prefixed by it. This option implies '--skip-empty'.",
+});
+
+co(function *() {
+    const args = parser.parseArgs();
+    const keepRegex = (null === args.keep_as_submodule) ?
+        null :
+        new RegExp(args.keep_as_submodule);
+    function keep(name) {
+        return null !== keepRegex && null !== keepRegex.exec(name);
+    }
+    const options = {
+        numParallel: args.j,
+        keepAsSubmodule: keep,
+        fetch: !args.no_fetch,
+        skipEmpty: args.skip_empty || (null !== args.root),
+    };
+    if (!args.no_fetch && null === args.url) {
+        console.error("URL is required unless '--no-fetch'");
+        process.exit(-1);
+    }
+    if (null !== args.url) {
+        options.url = args.url;
+    }
+    if (null !== args.root) {
+        console.log(`Joining from ${args.root}`);
+        options.joinRoot = args.root;
+    }
+    try {
+        yield StitchUtil.stitch(args.repo,
+                                args.commitish,
+                                args.target_branch,
+                                options);
+    }
+    catch (e) {
+        console.error(e.stack);
+        process.exit(-1);
+    }
+});
--- a/node/lib/util/stitch_util.js
+++ b/node/lib/util/stitch_util.js
@@ -1,0 +1,896 @@
+/*
+ * Copyright (c) 2017, Two Sigma Open Source
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of git-meta nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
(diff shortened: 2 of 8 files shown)


node/test/util/stitch_util.js
```diff
@@ -0,0 +1,833 @@
...
+    return sha.slice(0, 2) + sha.slice(3);
+}
+
+/**
+ *  - Replace `refs/stitched/converted/${splitSha(sha)}}` refs with their
+ *    equivalent logical mapping, e.g., if logical commit "1" maps to "aabb",
+ *    then we will remap `refs/stitched/converted/aa/bb` to
+ *    `refs/stitched/converted/1`.
+ */
+function refMapper(actual, mapping) {
```
novalis: I do not think refs are necessarily the right thing to use for this (I'm putting this note here because this is where I happened to notice it, but I'm not just talking about the tests).  Updating packed refs is expensive -- the entire file must be rewritten.   Notes make more sense to me, since they are designed for this, and can be added in more like O(log) time.

node/lib/util/stitch_util.js
```diff
@@ -0,0 +1,637 @@
...
+ *
+ * @param {String} sha
+ * @return {String}
+ */
+exports.convertedRefName = function (sha) {
+    return "refs/stitched/converted/" + exports.splitSha(sha);
+};
+
+/**
+ * Return the name of the ref indicating that the submodule commit having the
```
novalis: Wait, why do we need the submodule commit to be under the meta sha?

node/lib/util/stitch_util.js
```diff
@@ -0,0 +1,637 @@
...
+ * @return {String}
+ */
+exports.fetchedSubRefName = function (metaSha, subSha) {
+    return "refs/stitched/fetched/" + exports.splitSha(metaSha) + "/sub/" +
+        exports.splitSha(subSha);
+};
+
+/**
+ * Return a summary for the specified `commit` for the submodule having the
+ * specifiewd `name` to be used to describe changes that are rolled into a
```
novalis: spelling of specified.

node/lib/util/stitch_util.js
```diff
@@ -0,0 +1,637 @@
...
+Includes changes from submodule ${name} on ${subCommit.id().tostrS()}.
+Author: ${author.name()} <${author.email()}>
+Date:   ${Commit.formatCommitTime(author.when())}
+
+${subCommit.message()}
+`;
+};
+
+/**
+ * Return true if there is no difference in the specified `newTree` and
```
novalis: It's confusing to have "isEmpty" be some sort of equality comparison thing.

Also, an empty tree is illegal, but maybe we use it for some special thing.

node/lib/util/stitch_util.js
```diff
@@ -0,0 +1,637 @@
...
+ *
+ * @param {String} entry
+ * @param {Object} parentMap  from sha to Set of its direct parents
+ * @return {[String]}
+ */
+exports.listCommitsInOrder = function (entry, parentMap) {
+    assert.isString(entry);
+    assert.isObject(parentMap);
+
+    // First, compute the levels of the commits.  A level '0' means that a
```
novalis: FWIW, git internally uses the term "generation" for this (in some new code that I was looking at recently).

And I would say "a generation N means that the max generation of the commit's parents is (N - 1)", but maybe that's not actually clearer.

node/lib/util/stitch_util.js
```diff
@@ -0,0 +1,637 @@
...
+            }
+            else if (undefined !== level) {
+                // If all parents computed thus far, recompute the max.  It can
+                // not be less than or equal to any parent.
+
+                level = Math.max(level, parentLevel + 1);
+            }
+        }
+        if (undefined !== level) {
+            // We were ab le to compute it, store and pop.
```
novalis: spurious space in able

node/lib/stitch.js
```diff
@@ -0,0 +1,243 @@
...
+
+parser.addArgument(["--root"], {
+    type: "string",
+    help: "When provided, run a *join* operation that creates a new history \
+joining those of the submodules under the specified 'root' path.  \
+In this history those paths will be relative to 'root', i.e., they will not \
+be prefixed by it. This option implies '--skip-empty'.",
+});
+
+const stitch = co.wrap(function *(repoPath,
```
novalis: Should we wrap some of these args into an options object?

node/lib/util/stitch_util.js
```diff
@@ -0,0 +1,637 @@
...
+ * 
+ * - url -- submodule configured url
+ * - sha -- submodule sha
+ * - metaSha -- sha it was introcued on
+ *
+ * @param {NodeGit.Repository}  repo
+ * @param {String}              url
+ * @param {[Object]}            subFetches
+ */
+exports.fetchSubCommits = co.wrap(function *(repo, url, subFetches) {
```
novalis: This is, in general, super inefficient.  In 99% of cases, submodule commits will be fast-forward, so if we fetch the most-recent one we can then check to see if we have each remaining one before trying to fetch it.  And if we do have it, we don't need to write a synthetic ref for it, since that means that the synthetic ref written for the first thing we fetched will contain it.

node/lib/stitch.js
```diff
@@ -0,0 +1,243 @@
...
+    for (let i = 0; i < commitsToStitch.length; ++i) {
+        const next = commitsToStitch[i];
+
+        // If we had the same commit in the graph more than once, don't convert
+        // it more than once.
+
+        const nextSha = next.id().tostrS();
+        const parents = yield next.getParents();
+        const newParents = [];
+        for (let i = 0; i < parents.length; ++i) {
```
novalis: nit: shadowing the variable "i" (and also you could just use for (const parent of parents)

node/lib/stitch.js
```diff
@@ -0,0 +1,243 @@
...
+        yield DoWorkQueue.doInParallel(subNames, doFetch, numParallel);
+    }
+
+    console.log("Now stitching");
+    let lastCommit = null;
+
+    for (let i = 0; i < commitsToStitch.length; ++i) {
+        const next = commitsToStitch[i];
+
+        // If we had the same commit in the graph more than once, don't convert
```
novalis: This does not actually seem to check that right here?

node/lib/util/stitch_util.js
```diff
@@ -0,0 +1,637 @@
...
+
+    const stitchSub = co.wrap(function *(name, sha) {
+        let subCommit;
+        try {
+            subCommit = yield repo.getCommit(sha);
+        }
+        catch (e) {
+            console.error("On meta-commit", commit.id().tostrS(),
+                          name, "is missing", sha);
+            return;                                                   // RETURN
```
novalis: Do we want to allow this?

node/lib/stitch.js
```diff
@@ -0,0 +1,243 @@
+#!/usr/bin/env node
+/*
+ * Copyright (c) 2017, Two Sigma Open Source
```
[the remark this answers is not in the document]
bpeabody: I did author it last year.

node/test/util/stitch_util.js
bpeabody: Yes, I like it. Will change.

node/lib/stitch.js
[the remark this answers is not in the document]
bpeabody: I did author it last year.

node/lib/stitch.js
bpeabody: Yes, I think I'll do that and also move the function into `stitch_util` and give it a test driver.

node/lib/stitch.js
bpeabody: Yes, looks like that comment is dangling.

node/lib/stitch.js
bpeabody: will fix

node/test/util/stitch_util.js
novalis: (actually, the packed-refs file only needs a rewrite when things are changed -- new refs are added loose -- but I still think notes make more sense)

node/lib/util/stitch_util.js
```diff
@@ -0,0 +1,637 @@
...
+ */
+exports.splitSha = function (sha) {
+    const pre = sha.slice(0, 2);
+    const post = sha.slice(2);
+    return `${pre}/${post}`;
+};
+
+/**
+ * Return the name of the ref used to indicate that the commit having the
+ * specified `sha` has beren converted.
```
bpeabody: SP

node/lib/util/stitch_util.js
bpeabody: It's mapping from meta-repo commit sha to stitched commit sha

node/lib/util/stitch_util.js
bpeabody: Yes, that's misleading. It should be something like `isUnchanged`.

node/lib/util/stitch_util.js
bpeabody: I'll use the git term (In the past I've used the term "levalization")

node/lib/util/stitch_util.js
bpeabody: `GitUtil.fetchSha` checks to see if the commit already exists before attempting to fetch it:

```JavaScript
/**
 * Fetch the specified `sha` from the specified `url` into the specified
 * `repo`, if it does not already exist in `repo`.
 *
 * @async
 * @param {String}             url
 * @param {String}             sha
 */
exports.fetchSha  = co.wrap(function *(repo, url, sha) {
```

node/lib/util/stitch_util.js
novalis: No it isn't.

bpeabody: OK, I (originally) thought you were talking about `convertedRefName`, but you're referring to next function, `fetchedSubRefName`. What I'm doing here is making a ref for submodule commits I fetch so that they don't get GC'd. I'll make this more clear in the comments.

novalis: Couldn't we just use an ordinary synthetic ref?

node/lib/util/stitch_util.js
novalis: Can we make this run in most-recent-first order?  And do we want to avoid writing synthetic refs for shas that we didn't have to fetch?

node/lib/util/stitch_util.js
bpeabody: Yes, that makes sense.

node/lib/util/stitch_util.js
bpeabody: I do fetch them in that order (otherwise it would take months or years to run), but I'm making the synthetic refs whether a fetch is needed or not. I can change this to not make the ref unless a fetch happens (might need a tweak in the api for `GitUtil.fetchSha` to return a bool).

bpeabody:
By the way, @novalis, I've been pushing intermediate commits as I fix things, but I'm not quite done yet. Feel free to keep looking as I push if you want, but I'll ping you when I'm completely finished.

bpeabody:
Also, @novalis, I noticed a strange thing -- the git CL tool seems to refuse to recognize the notes I create with libgit2. If I fun `git notes show $ref` it says no notes found. If I run `git log --show-notes`, nothing shows up. Yet, they're there, and libgit2 finds them just fine.

bpeabody:
ah, yes. and `--ref refs/notes/whatever` for the `git-notes` command

bpeabody:
OK, @novalis, I'm ready for another look.

bpeabody:
Another push -- readied some code I'd accidentally removed from `test_util.js` to auto-cleanup temp files and a breathing test for the `StitchUtil.stitch` method.

node/lib/util/stitch_util.js
```diff
@@ -0,0 +1,875 @@
...
+    // So that we don't have to continuously re-read the `.gitmodules` file, we
+    // will assume that submodule URLs never change.
+
+    const getUrl = co.wrap(function *(commit, sub) {
+        let subUrl = urls[sub];
+
+        // If we don't have the url for this submodule, load them.
+
+        if (undefined === subUrl) {
+            console.log("loading urls");
```
novalis: Do we mean to keep this output?

node/lib/util/stitch_util.js
```diff
@@ -0,0 +1,875 @@
...
+ * @param {NodeGit.Repository}  repo
+ * @param {String}              url
+ * @param {[Object]}            subFetches
+ */
+exports.fetchSubCommits = co.wrap(function *(repo, url, subFetches) {
+    assert.instanceOf(repo, NodeGit.Repository);
+    assert.isString(url);
+    assert.isArray(subFetches);
+
+    for (let i = 0; i < subFetches.length; ++i) {
```
novalis: for/of subFetches?

node/lib/util/stitch_util.js
```diff
@@ -0,0 +1,875 @@
...
+        // Exit if we've already computed this one; can happen if one gets into
+        // the queue more than once.
+
+        if (next in generations) {
+            queue.pop();
+            continue;                                               // CONTINUE
+        }
+        let generation = 0;
+        const parents = parentMap[next] || [];
+        for (let i = 0; i < parents.length; ++i) {
```
novalis: Actually, there are a few for loops that should be for/of -- maybe it's worth grepping the diff  to check them all?

node/lib/stitch.js
```diff
@@ -0,0 +1,148 @@
...
+
+parser.addArgument(["--no-fetch"], {
+    required: false,
+    action: "storeConst",
+    constant: true,
+    defaultValue: false,
+    help: `If provided, assume commits are present and do not fetch.`,
+});
+
+parser.addArgument(["-t", "--target-branch"], {
```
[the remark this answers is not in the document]
bpeabody: Right now the `--target-branch` describes where to write the result of the stitch. I guess we can revisit this if we find it convenient to operate in a different way.

node/lib/util/stitch_util.js
```diff
@@ -0,0 +1,875 @@
...
+    assert.isString(originalSha);
+    const note = yield GitUtil.readNote(repo,
+                                        exports.convertedNoteRef,
+                                        originalSha);
+    if (null === note) {
+        // No note exists, return undefined to indicate no conversion happened.
+        return undefined;                                             // RETURN
+    }
+    const message = note.message();
+    return "" === message ? null : message;
```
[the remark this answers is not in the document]
bpeabody: `Note.message()` won't return null.

node/lib/util/stitch_util.js
bpeabody: Yes, I'll see if I can stomp them all. Need to get out of this habit.

node/lib/util/stitch_util.js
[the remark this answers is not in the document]
novalis: Sorry, I mean returns empty

bpeabody:
I set it to empty to indicate that the commit has been processed and cannot be stitched — e.g. an empty commit with no parent when skipping empty commits.

novalis:
Ah, legit.

node/lib/util/stitch_util.js
bpeabody: No, I don't think this is useful output.

node/lib/util/stitch_util.js
```diff
@@ -0,0 +1,875 @@
...
+    function compareCommits(a, b) {
+        const aGeneration = generations[a];
+        const bGeneration = generations[b];
+        if (aGeneration !== bGeneration) {
+            return aGeneration - bGeneration;                         // RETURN
+        }
+        if (a < b) {
+            return -1;                                                // RETURN
+        }
+        return 1;
```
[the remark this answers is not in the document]
bpeabody: This one is for ordering keys from a map; they can't be equal. I added a comment and changed to use the ternary operator.

bpeabody:
I think I addressed everything; ready for another look.

bpeabody:
(@novalis )

node/lib/stitch.js
```diff
@@ -0,0 +1,148 @@
...
+    required: false,
+    defaultValue: null,
+});
+
+parser.addArgument(["--skip-empty"], {
+    required: false,
+    action: "storeConst",
+    constant: true,
+    defaultValue: false,
+    help: "Skip a commit if its  tree would be the same as its first parent.",
```
novalis: nit: extra space after its

PR #662, opened 2018-09-17
Add index flag to stash and set default behavior to not use it

blevz:


novalis:
Oh, that makes sense.

blevz:
Squashed

PR #664, opened 2018-09-20
stitch optimizations and cleanup

bpeabody:


node/lib/util/stitch_util.js
```diff
@@ -52,41 +53,124 @@ const FILEMODE            = NodeGit.TreeEntry.FILEMODE;
...
+        currentTree = yield currentCommit.getTree();
+    }
+    const odb = yield repo.odb();
+    const changes = {};
+    const ODB_BLOB = 3;
+    const BLOB = NodeGit.TreeEntry.FILEMODE.BLOB;
+    const writeBlob = co.wrap(function *(sha) {
+        const content = contents[sha];
+        const blobId = yield odb.write(content, content.length, ODB_BLOB);
+        changes[sha] = new TreeUtil.Change(blobId, BLOB);
```
novalis: That is not how notes trees are supposed to work -- there's a whole sharding scheme.  Isn't there something in NodeGit that does this?

bpeabody: Git seems to be happy with it in that I can read notes written this way with `git notes show`. However, you're right, I should change to use the sharding scheme.

libgit2 does not provide a way to read or write more than one note at a time; these bulk operations provided a significant performance improvement.

PR #668, opened 2018-09-27
Better logic to compute which submodules to push

bpeabody:
The previous logic was to rule out pushing submodules if their current
commits (or their descendants) existed in one of a set of relevant
branches (e.g., tracking branches).  The downside of this strategy was
that we couldn't deduce parent-child relationships of submodule commits
that didn't exist locally.

The new strategy is to include as candidates only submodules that were
not changed between the commit being pushed and any of the (merge bases
of the) candidate branches.

Also fixed a bug where we were not constructing the ref name of a
tracking branch correctly.

### DIFF
--- a/node/lib/util/push.js
+++ b/node/lib/util/push.js
@@ -51,21 +51,12 @@ const UserError           = require("./user_error");
  * For a given proposed push, return a map from submodule to sha,
  * excluding any submodules that the server likely already has.
  *
- * Start with the set of submodules that are either open, or that
- * exist in .git/modules. Populate a pushMap from submodule name to
- * commit (SHA) registered in the meta-commit being pushed.
+ * Create a list of relevant branches, including any tracking branches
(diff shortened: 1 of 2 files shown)


node/lib/util/push.js
```diff
@@ -51,21 +51,12 @@ const UserError           = require("./user_error");
...
- * in the target.
- *
- * Note that for "exists" we mean the target points to exactly the
- * commit needed or a descendant of that commit.
- *
- * If the user is pushing a branch, B, further reduce the size of
- * pushMap by removing commits referenced in the branches that B pulls
- * from and/or pushes to by default, if they exist and are different
- * from the target branch.
+ * Push return in the map only those submodules that (1) exist locally, in
```
novalis: grammar: "Push return" seems confusing.

bpeabody:
Fixed that wording and will merge.

PR #671, opened 2018-10-04
Use normal routine to find repo for destitch

bpeabody:
The normal git-meta way isn't needed (we're not in a meta-repo) and
doesn't work if you're using worktrees.


bpeabody:
Well, I'm going to need to fix `git-meta-root` to discern the difference in submodule gitlinks, but this is still the right thing to do.

novalis:
You've got a jshint failure in CI

bpeabody:
Whoops, will fix

PR #675, opened 2018-10-16
Module import optimizations to git-meta

jhedwardyang:
**NodeGit is slow.** 
These module import optimizations result in some improved performance in specific targeted areas where interactive performance is useful. 

Performance changes (mid-2012 MBA w/ SSD)
* **git meta root, 700ms -> 260ms (-63%)**
* --help (and other strictly argparse-only operations), 650ms -> 230ms (-65%)
* forwarded commands to pure-git all improved. eg.
* * log 650ms -> 300ms (-54%)
* * diff 700ms -> 270ms (-62%)

_Highly recommend reviewing this commit-by-commit._

### DIFF
--- a/node/lib/git-meta.js
+++ b/node/lib/git-meta.js
@@ -153,6 +153,7 @@ const commands = {
 
 // These optimized commands do not require NodeGit, and can skip importing it.
 const optimized = {
+    "root": true,
     "version": true,
 };
 
--- a/node/lib/git-meta.js
+++ b/node/lib/git-meta.js
@@ -69,11 +69,9 @@ const version      = require("./cmd/version");
  * @param {ArgumentParser} parser
  * @param {String}         commandName
  * @param {Object}         module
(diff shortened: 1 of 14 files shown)


node/lib/git-meta.js
```diff
@@ -63,30 +62,35 @@ const submodule    = require("./cmd/submodule");
...
         func: function (args) {
+            // This optimization allows to skip importing NodeGit
+            // if not required (eg. version).
+            if (!skipNodeGit) {
+                const NodeGit = require("nodegit");
+                // see https://github.com/nodegit/nodegit/issues/827 -- this is
+                // required to prevent random hard crashes with e.g.
+                // parallelism  in index operations. Eventually, this will be
+                // nodegit's default.
+                NodeGit.setThreadSafetyStatus(
```
bpeabody: @novalis -- can we skip this now?

Issue #678, opened 2018-10-19
empty commits can be created when cherry-picking or rebasing


shijinglu:
One way to reproduce this issue is to cherry-picking the same commit over and over again. For instance if current git tree is like the following:

```
* 141f8bd - (25 hours ago) m2 - Shijing Lu (HEAD -> master, origin/master, origin/HEAD)
* a0ff720 - (25 hours ago) m1 - Shijing Lu
| * c9d9c8d - (26 hours ago) f2 - Shijing Lu (feature)
| * 1ffe97e - (26 hours ago) f1- Shijing Lu
| * b3721fe - (26 hours ago) f0 - Shijing Lu
|/
* c1b7788 - (26 hours ago) Initial commit - shijing

```
and if I keep running the following command:
`$ git meta cherry-pick feature`

then git meta will create a row of empty commits stacking on each other like 

```
* 60ed21f - (3 weeks ago) f2 - Shijing Lu
* 4b1b9f1 - (3 weeks ago) f2 - Shijing Lu
* 7c6be52 - (3 weeks ago) f2 - Shijing Lu
* 8a72c30 - (3 weeks ago) f2 - Shijing Lu
* fa8e1c8 - (3 weeks ago) f2 - Shijing Lu
* 141f8bd - (25 hours ago) m2 - Shijing Lu (HEAD -> master, origin/master, origin/HEAD)
* a0ff720 - (25 hours ago) m1 - Shijing Lu
| * c9d9c8d - (26 hours ago) f2 - Shijing Lu (feature)
| * 1ffe97e - (26 hours ago) f1- Shijing Lu
| * b3721fe - (26 hours ago) f0 - Shijing Lu
|/
* c1b7788 - (26 hours ago) Initial commit - shijing

```

This isn't a desired behavior as empty commits often causes issue in vanilla git.

PR #679, opened 2018-10-19
#678 fix git meta creates empty commits issue

shijinglu:
git-meta calculates the submodule changes by checking if the submodule changes map is empty:
` 0 !== Object.keys(picks.commits).length)` 

However, there are cases (check the issue ticket for details), the commits map is not empty but is equivalent to empty because the changes within each submodule has an empty change. aka it is like the following:

```json
"commits": {
      "submodule_a": {}
 }
```

There are generally two ways to fix this, one way is to make sure submodules never report empty commits or have cherry_pick_utils smart enough not to create a new meta commit if the commit is effectively empty. 

I choose the later because the check was already in the code, this PR only makes it cover the corner cases.

### DIFF
--- a/node/lib/util/cherry_pick_util.js
+++ b/node/lib/util/cherry_pick_util.js
@@ -550,7 +550,7 @@ exports.rewriteCommit = co.wrap(function *(repo, commit) {
     const nChanges = Object.keys(picks.commits)
         .map(name => Object.keys(
             picks.commits[name]).length + picks.ffwds[name] ? 1 : 0
-        ).reduce((acc, len) => acc+len, 0);
+        ).reduce((acc, len) => acc + len, 0);
         if ("" === errorMessage &&
         (0 !== Object.keys(changes.simpleChanges).length || 0 !== nChanges)) {
         result.newMetaCommit =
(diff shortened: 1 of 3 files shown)


node/lib/util/cherry_pick_util.js
```diff
@@ -544,9 +547,12 @@ exports.rewriteCommit = co.wrap(function *(repo, commit) {
         newMetaCommit: null,
     };
     yield SparseCheckoutUtil.writeMetaIndex(repo, index);
-    if ("" === errorMessage &&
-        (0 !== Object.keys(changes.simpleChanges).length ||
-                                    0 !== Object.keys(picks.commits).length)) {
+    const nChanges = Object.keys(picks.commits)
+        .map(name => Object.keys(
+            picks.commits[name]).length + picks.ffwds[name] ? 1 : 0
+        ).reduce((acc, len) => acc+len, 0);
```
novalis: nit: spaces around operators

node/test/util/cherry_pick_util.js
```diff
@@ -834,6 +841,10 @@ Submodule ${colors.red("s")} is conflicted.
             const eightCommit = yield x.getCommit(eightCommitSha);
             const result  = yield CherryPickUtil.cherryPick(x, eightCommit);
 
+            if(c.duplicate) {
```
novalis: nit: space before (

novalis:
Cool.  I usually squash my commits before merging, so please do that and then merge it.

shijinglu:
> Cool. I usually squash my commits before merging, so please do that and then merge it.

I don't think I have the permission to merge

PR #682, opened 2018-10-23
Optimize Push logic to compare against remote refs to determine what to push

jhedwardyang:
This code optimizes the push performance for new refs (without tracking) by finding a commit which has not yet been pushed, and parents have been pushed. Then comparing the pushed commit against that commit's parent commit to determine what to push.

### DIFF
--- a/node/lib/util/git_util.js
+++ b/node/lib/util/git_util.js
@@ -44,7 +44,6 @@ const NodeGit      = require("nodegit");
 const path         = require("path");
 
 const ConfigUtil          = require("./config_util");
-const DoWorkQueue         = require("./do_work_queue");
 const Hook                = require("./hook");
 const GitUtilFast         = require("./git_util_fast");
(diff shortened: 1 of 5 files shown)


novalis:
I am worried about the case of a merge where both sides got an additional commit before the merge. There are thus three new commits to push, but their common ancestor should not be pushed. Can you describe how your code handles this situation?

On October 23, 2018 5:57:02 PM EDT, Edward Yang <notifications@github.com> wrote:
>@jhedwardyang requested your review on: twosigma/git-meta#682 Optimize
>Push logic to compare against remote refs to determine what to push.
>
>-- 
>You are receiving this because your review was requested.
>https://github.com/twosigma/git-meta/pull/682#event-1921848275

-- 

bpeabody:
@novalis how about we make sure there's a test case that validates the expected behavior in this situation?

PR #687, opened 2018-11-09
Fix git-meta rm path overmatching issue

shijinglu:
Currently running `git meta rm -rf abc` matches following directories which should not have:

```
abcd/
abcd/api
abcd/service
...
```

The issue is because `getSubmodulesInPath` in the `submodule_utils.js` has a bug: it thinks if x.startsWith(y) than x must be a subdirectory of y. This is not true as `abcd/ef/` starts  with `abc`, but it is not a subdirectory of the later.

I changed to logic to:
1. If y has a trailing '/' then `startsWith` is sufficient
2. Otherwise, removing prefix `x` from y, the rest of y should starts with a leading `/`.

### DIFF
--- a/node/lib/util/merge_util.js
+++ b/node/lib/util/merge_util.js
@@ -38,7 +38,6 @@ const NodeGit      = require("nodegit");
 
 const Checkout            = require("./checkout");
 const CherryPickUtil      = require("./cherry_pick_util");
-const Commit              = require("./commit");
 const ConfigUtil          = require("./config_util");
 const DoWorkQueue         = require("./do_work_queue");
 const GitUtil             = require("./git_util");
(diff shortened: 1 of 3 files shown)


node/lib/util/submodule_util.js
```diff
@@ -559,11 +559,19 @@ exports.getSubmodulesInPath = function (dir, indexSubNames) {
     if ("" === dir) {
         return indexSubNames;
     }
+
+    // test if the short path a parent dir of the long path
+    const isParentDir = (short, long) => {
+        return long.startsWith(short) && (
+            short[short.length-1] === "/" ||
+            long.replace(short, "")[0] === "/"
```
novalis: this check could be long[short.length] === '/', right?

shijinglu: yah, logically they are the same, but yours is conciser!

node/lib/util/merge_util.js
```diff
@@ -38,7 +38,6 @@ const NodeGit      = require("nodegit");
 
 const Checkout            = require("./checkout");
 const CherryPickUtil      = require("./cherry_pick_util");
-const Commit              = require("./commit");
```
shijinglu: removed according to jslint

PR #690, opened 2019-02-01
Allow normal merge to operate on bare and half opened repos

shijinglu:
Add an option to `git meta merge` to allow it choose working directory mode:
1. open submodule is not allowed
2. always fully open submodule
3. open submodule only when manually conflict resolution is required.

Also refactor the existing `merge` module to make it less monolithic


node/lib/util/merge_util.js
```diff
@@ -129,294 +141,518 @@ exports.fastForwardMerge = co.wrap(function *(repo, commit, message) {
...
+});
 
-        const fromSha = change.newSha;
-        yield fetcher.fetchSha(subRepo, name, fromSha);
-        const subHead = yield subRepo.getHeadCommit();
-        const headSha = subHead.id().tostrS();
-        const fromCommit = yield subRepo.getCommit(fromSha);
+/**
+ * Write tree representation of the index to the disk, create a commit
+ * from the tree and update reference if needed.
```
novalis: So, this is sort of odd.  Why not have the merge command update the ref (and the merge-bare command not)?  That is, try to avoid having a method that does two different things.

node/lib/util/status_util.js
```diff
@@ -613,3 +613,36 @@ Before proceeding, you must complete the ${command} in progress (by running
 'git meta ${command} --abort').`);
     }
 };
+
+/**
+ * Return an error message if the specified `status` of a repository isn't in a
+ * normal, ready state, that is, it does not have any conflicts or in-progress
```
novalis: This is hard to parse.  There are too many negatives.  Can we rephrase?

node/lib/util/status_util.js
```diff
@@ -613,3 +613,36 @@ Before proceeding, you must complete the ${command} in progress (by running
...
+ * operations from the sequencer.  Adjust output paths to be relative to the
```
novalis: I have no idea what the "adjust output paths" sentence means but it doesn't seem to relate to anything in this function?

node/lib/util/open.js
```diff
@@ -267,6 +300,9 @@ Opener.prototype.getSubrepo = co.wrap(function *(subName, bare) {
                                                  true);
         }
     }
+    else if(this.d_absorbedSubs.has(subName) && WDTYPE.ALLOW_BARE === wdType) {
```
novalis: nit: space before (

node/lib/util/open.js
```diff
@@ -237,26 +260,36 @@ Opener.prototype.isAtLeastHalfOpen = co.wrap(function *(subName) {
 
 /**
  * Return the repository for the specified `submoduleName`, opening it if
- * necessary.
+ * necessary based on the expected working directory type:
+ *  1. FORCE_BARE
+ *      - directly return opened absored sub if there is one
```
novalis: spelling of "absorbed"

bpeabody:
If we have this capability, I don't see any reason to expose the options to end users; we should open submodules only when absolutely necessary.

node/lib/util/merge_common.js
```diff
@@ -0,0 +1,317 @@
...
+    return this.d_ourCommit;
+});
+
+/**
+ * return an index object that contains the merge changes and whose tree
+ * representation will be flushed to disk.
+ * @async
+ * @return {NodeGit.Index} 
+ */
+MergeContext.prototype.getIndexToWrite = co.wrap(function *(){
```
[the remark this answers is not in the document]
shijinglu: In the bare mode, the `index` comes from  [`Merge.commits`](https://libgit2.org/libgit2/#HEAD/group/merge/git_merge_commits), because we cannot assume there is an index object in this mode. 

In the interactive mode, the `index` object is read from the current repo, and then we can add edited files later with [`index.addByPath(...)`](https://github.com/twosigma/git-meta/blob/master/node/lib/util/merge_util.js#L230)

bpeabody:
Also, please get @harikmenon to sign of on any UX changes.

node/lib/util/merge_util.js
shijinglu: Now it is the case that interactive merge update the ref ("HEAD") and the bare merge not. But interactive merge does more than that: 
1. it operates directly on repo's current index, so users can see all the file changes if merge is paused for user actions.
2. it pulls necessary changes when needed
3. it force reset subrepo's head when needed

node/lib/util/merge_common.js
```diff
@@ -0,0 +1,317 @@
...
+ * 
+ * @async
+ * @return {String} commit message
+ */
+MergeContext.prototype.getCommitMessage = co.wrap(function *(){
+    const message = (null === this.d_commitMessage) ? 
+        GitUtil.stripMessage(yield this.d_editMessage()) :
+        this.d_commitMessage;
+    if ("" === message) {
+        console.log("Empty commit message.");
```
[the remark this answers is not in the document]
shijinglu: Actually I struggled a little here. I wanted to fail the merge here, but the current merge does not fail the merge. Then I am sure if this is a feature or a potential bug ...

node/lib/util/merge_common.js
```diff
@@ -0,0 +1,317 @@
...
+ * @returns {SubmoduleFetcher}
+ */
+MergeContext.prototype.getFetcher = co.wrap(function *(){
+    return yield this.d_opener.fetcher();
+});
+
+exports.MergeContext = MergeContext;
+
+/**
+ * A class that tracks result from mergeing steps.
```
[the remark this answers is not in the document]
shijinglu: ack

node/lib/util/merge_common.js
```diff
@@ -0,0 +1,317 @@
...
+     * @property {Stirng|null}
+     */
+    get errorMessage() {
+        return this.d_errorMessage;
+    }
+
+    /**
+     * @property {String|null}
+     */
+    get infoMessage(){
```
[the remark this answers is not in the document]
shijinglu: ack

node/lib/util/merge_common.js
```diff
@@ -0,0 +1,317 @@
...
+
+    /**
+     * @property {String|null}
+     */
+    get finishSha() {
+        return this.d_finishSha;
+    }
+
+    /**
+     * @property {string[]}
```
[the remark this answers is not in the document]
shijinglu: ahh, wrong jsdoc

node/lib/util/open.js
```diff
@@ -43,6 +43,17 @@ const SubmoduleUtil       = require("./submodule_util");
 const SubmoduleConfigUtil = require("./submodule_config_util");
 const SubmoduleFetcher    = require("./submodule_fetcher");
 
+/**
+ * @enum {WDTYPE}
+ * Flags that describe what working directory type the repo is.
```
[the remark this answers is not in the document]
shijinglu: right, if we "always" open submodule is not necessary, we don't even need an enum here.

node/lib/util/open.js
shijinglu: ack

node/lib/util/open.js
shijinglu: 👍 , also updated all the `){`

node/lib/util/status_util.js
shijinglu: interesting, is this a github bug, because I don't think I made any changes to [this file](https://github.com/twosigma/git-meta/blob/master/node/lib/util/status_util.js#L585-L592)

node/lib/util/status_util.js
shijinglu: me neither, think this patch will go away after rebase

novalis:
On February 1, 2019 10:02:51 AM EST, shijing <notifications@github.com> wrote:
>shijinglu commented on this pull request.
>
>
>
>> +    return this.d_changeIndex;
>+});
>+
>+/**
>+ * Return the previously set/built commit message, or use the callback
>to
>+ * build commit messsage. Once built, the commit message will be
>cached.
>+ * 
>+ * @async
>+ * @return {String} commit message
>+ */
>+MergeContext.prototype.getCommitMessage = co.wrap(function *(){
>+    const message = (null === this.d_commitMessage) ? 
>+        GitUtil.stripMessage(yield this.d_editMessage()) :
>+        this.d_commitMessage;
>+    if ("" === message) {
>+        console.log("Empty commit message.");
>
>Actually I struggled a little here. I wanted to fail the merge here,
>but the current merge does not fail the merge. Then I am sure if this
>is a feature or a potential bug ...

Whatever plain git does, we should do.
-- 

node/lib/cmd/merge.js
```diff
@@ -91,6 +91,13 @@ exports.configureParser = function (parser) {
         constant: true,
         help: "abort an in-progress merge",
     });
+    parser.addArgument(["--open-submodule"], {
```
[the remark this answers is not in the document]
shijinglu: actually, there is no need to expose bare merge mode for end users. I am going to remove the option and make `allow_bare` the default option.

node/lib/util/merge_common.js
```diff
@@ -0,0 +1,317 @@
...
+                yield this.getChangeIndex(),
+                yield this.getOurCommit(),
+                this.d_theirCommit);
+    }
+    return this.d_changes;
+});
+
+/** 
+ * @async
+ * @return {NodeGit.Commit} return left side merge commit
```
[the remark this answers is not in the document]
shijinglu: we need to determine if `ourCommit` comes from direct user input or from "HEAD", we can either determine that at the very beginning or on the fly when needed. I choose the later case because I think being lazy is always good.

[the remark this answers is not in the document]
novalis: It's best to have separation of concerns.  One piece of code is responsible for figuring out what's going to be merged.  Another is responsible for doing the merge.

node/lib/util/merge_util.js
novalis: The first is just a question of what index is used, right?

The second of these should be true for both kinds of merge, right?  Unless by "pull", you really mean "pull" as in fetch+merge.  

Even if 1 and 2 do really both need to be conflated into this method, it does not follow that 3 does.  Our goal is clean code, which means separation of concerns as much as possible.

node/lib/util/merge_common.js
```diff
@@ -0,0 +1,317 @@
...
+     */
+    constructor(infoMessage, errorMessage, finishSha, submoduleCommits) {
+        this.d_infoMessage = infoMessage;
+        this.d_errorMessage = errorMessage;
+        this.d_finishSha = finishSha;
+        this.d_submoduleCommits = submoduleCommits;
+    }
+
+    /**
+     * @property {Stirng|null}
```
novalis: Spelling of String

node/lib/util/merge_util.js
```diff
@@ -129,294 +141,522 @@ exports.fastForwardMerge = co.wrap(function *(repo, commit, message) {
...
- * Merge the specified `subs` in the specified `repo` having the specified
- * `index`.  Stage submodule commits in `metaRepo`.  Return an object
- * describing any commits that were generated and conflicted commits.  Use the
- * specified `opener` to acces submodule repos.  Use the specified `message` to
- * write commit messages.
- * @param {NodeGit.Repository} metaRepo
- * @param {Open.Opener}        opener
- * @param {NodeGit.Index}      metaIndex
- * @param {Object}             subs        map from name to SubmoduleChange
+ * Get the left side commit and sha for submodule `subRepo` based on workding
```
novalis: spelling of working

novalis: but also "based on working directory type" doesn't really make a lot of sense to me.

node/lib/util/merge_util.js
```diff
@@ -129,294 +141,522 @@ exports.fastForwardMerge = co.wrap(function *(repo, commit, message) {
...
+ * directory type.
+ * 
+ * @async
+ * @param {NodeGit.Repository}  subRepo
+ * @param {SubmoduleChange}     subChange
+ * @param {SUB_OPEN_OPTION}     openOption
```
novalis: Like, this isn't really about our opening policy -- it's about what we're merging.  There are two options: (a) our left side is our current state (working tree / index), vs (b) our left side is an arbitrary sha.

node/lib/util/open.js
```diff
@@ -43,6 +43,17 @@ const SubmoduleUtil       = require("./submodule_util");
 const SubmoduleConfigUtil = require("./submodule_config_util");
 const SubmoduleFetcher    = require("./submodule_fetcher");
 
+/**
+ * @enum {SUB_OPEN_OPTION}
+ * Flags that describe wheter to open a submodule if it is part of a merge.
```
novalis: Spelling of whether

node/lib/util/open.js
```diff
@@ -43,6 +43,17 @@ const SubmoduleUtil       = require("./submodule_util");
...
+ */
+const SUB_OPEN_OPTION = {
+    FORCE_OPEN    : 0, // non-bare repo and open submodule when needed
```
novalis: "when needed" and "force" seem like an odd fit.  Also, do we actually want this ever?

node/lib/util/merge_util.js
```diff
@@ -129,294 +141,522 @@ exports.fastForwardMerge = co.wrap(function *(repo, commit, message) {
...
+        submoduleCommits: {},
+        errorMessage: null,
+    };
+    const mergeAsyncSteps = [
+        mergeStepPrepare,
+        mergeStepFF,
+        mergeStepMergeSubmodules,
+    ];
+
+    for (let i = 0; i < mergeAsyncSteps.length; i++) {
```
novalis: for (const asyncStep of asyncSteps)

node/lib/util/merge_common.js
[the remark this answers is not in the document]
shijinglu: yes, `getOurCommit` does exactly that. It is a helper that determines check if left side commit is given, and if not given use the `HEAD` if the we are not merging in a bare repo. 

I think this logic is unavoidable, the question is where we put this logic. IMO, the context object is a good place because it manages the objects that are used in the merge and `ourCommit` is one of the managed objects.

node/lib/util/merge_common.js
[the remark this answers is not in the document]
shijinglu: Later in the merge process, we do abort the merge if message is empty, just like vanilla git will do. But not sure if it was treated as an error internally in git. I don't think there are any differences

node/lib/util/merge_common.js
shijinglu: ack

node/lib/util/open.js
shijinglu: ack

node/lib/util/open.js
shijinglu: yes, we still need this option for at least two reasons:
1. `FORCE_OPEN` means "open the submodule when creating the `subRep` object". And creating the `subRepo` object is called in many places, not just the merge. For example the `subRepo` objects are needed in `reset`, `stash`, `commit` etc.  It is hard to test if half open submodules work for all those case. So to be safe let's allow fully opening in other commands.
2. Now the default option for normal merge is `ALLOW_BARE`, but in case there are bugs in its workflow, we can quickly switch back to old merge workflow with `FORCE_OPEN` instead of going through painful git-revert process.

novalis:
On February 5, 2019 5:57:25 PM EST, shijing <notifications@github.com> wrote:
>shijinglu commented on this pull request.
>
>
>
>> @@ -43,6 +43,17 @@ const SubmoduleUtil       =
>require("./submodule_util");
> const SubmoduleConfigUtil = require("./submodule_config_util");
> const SubmoduleFetcher    = require("./submodule_fetcher");
> 
>+/**
>+ * @enum {SUB_OPEN_OPTION}
>+ * Flags that describe wheter to open a submodule if it is part of a
>merge.
>+ */
>+const SUB_OPEN_OPTION = {
>+    FORCE_OPEN    : 0, // non-bare repo and open submodule when needed
>
>yes, we still need this option for at least two reasons:
>1. `FORCE_OPEN` means "open the submodule when creating the `subRep`
>object". And creating the `subRepo` object is called in many places,
>not just the merge. For example the `subRepo` objects are needed in
>`reset`, `stash`, `commit` etc.  It is hard to test if half open
>submodules work for all those case. So to be safe let's allow fully
>opening in other commands.

I don't think any of those three cases could possibly require opening a submodule. Unless you mean "stash apply"? 

>2. Now the default option for normal merge is `ALLOW_BARE`, but in case
>there are bugs in its workflow, we can quickly switch back to old merge
>workflow with `FORCE_OPEN` instead of going through painful git-revert
>process.

I would rather revert, since this change is pretty hairy, and there's no reason to believe that just changing this option would in fact restore the old behavior.

-- 

node/lib/cmd/merge.js
```diff
@@ -167,8 +176,24 @@ Merge of '${commitName}'
         return GitUtil.editMessage(repo, message);
     };
     const commit = yield repo.getCommit(commitish.id());
-    const result =
-          yield MergeUtil.merge(repo, commit, mode, args.message, editMessage);
+    let wdType;
+    switch (args.open_submodule.toLowerCase()) {
```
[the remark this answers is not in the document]
shijinglu: switch statement removed

node/lib/cmd/merge_bare.js
```diff
@@ -62,12 +62,12 @@ exports.configureParser = function (parser) {
...
         type: "string",
         help: "our side commitish to merge",
-        nargs: 1,
+        nargs: "?",
     });
     parser.addArgument(["theirCommit"], {
         type: "string",
         help: "their side commitish to merge",
-        nargs: 1,
+        nargs: "?",
```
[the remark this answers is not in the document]
shijinglu: updated

node/lib/util/merge_util.js
shijinglu: Ok, let's move the logic to `MergeContext` then.

node/lib/util/merge_util.js
shijinglu: actually, after revisiting the compute changes logic, the left side commit for submodules have already computed correctly in `CherryPickUtil.computeChangesBetweenTwoCommits`, so let me remove this

node/lib/util/merge_util.js
shijinglu: ack

shijinglu:
> I don't think any of those three cases could possibly require opening a submodule. Unless you mean "stash apply"?

yah, stash apply opens the submodule, also `git meta amend` opens the submodule as well.  cherry-pick-abort when it resets the HEAD to some commit, it also opens the submodule

node/lib/util/merge_util.js
shijinglu: correct. So basically the logic is now like this: 
1. `merge` command sets the `ourCommit` to null and `theirCommit` equal to a sha in the `MergeContext`. And `MergeContext` then set `ourCommit` to HEAD if current repo is not meata, and raises an error if it is.
2. `merge-bare` sets both `ourCommit` and `theirCommit`
3. left side commits of the merge for each submodule are also computed in the `MergeContext` based on the meta repo merging commits

novalis:
> > I don't think any of those three cases could possibly require opening a submodule. Unless you mean "stash apply"?
> 
> yah, stash apply opens the submodule, also `git meta amend` opens the submodule as well. cherry-pick-abort when it resets the HEAD to some commit, it also opens the submodule

(not sure github is going to put this comment in the right place, but why would commit --amend need to open a submodule?  And why would cherry-pick --abort need to do so?

shijinglu:
> (not sure github is going to put this comment in the right place, but why would commit --amend need to open a submodule? And why would cherry-pick --abort need to do so?

But the two are know situations and our users often rely on `git meta commit --amend` to open submodules.  Is opening submodules necessary, may be may be now. But I think it is not up to this PR to answer these two questions.  Let's keep these behaviors and if we feel necessary, we can always come back and remove `FORCE_OPEN` later.

node/lib/util/merge_common.js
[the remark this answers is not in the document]
novalis: Then maybe we should log that message later too?

node/lib/util/merge_util.js
```diff
@@ -129,294 +141,492 @@ exports.fastForwardMerge = co.wrap(function *(repo, commit, message) {
...
+ * 
+ * @async
+ * @param {NodeGit.Index}      metaIndex index of the meta repo
+ * @param {String}             subName submodule name
+ * @param {SubmoduleChange}    change specifies the commits to merge
+ * @param {String}             message commit message
+ * @param {SubmoduleFetcher}   fetcher helper to fetch commits in the sub
+ * @param {NodeGit.Signature}  sig default signature
+ * @param {Open.Opener}        opener helper to open a sub
+ * @param {SUB_OPEN_OPTION}    openOption opention to open a sub
```
novalis: spelling of option.

node/test/util/merge_bare.js
```diff
@@ -111,7 +113,7 @@ x=U:C3-2 s=Sa:b;C4-2 s=Sa:a;Bmaster=3;Bfoo=4`,
                 theirCommit: "4",
                 ourCommit: "3",
             },
-            "non-ffmerge with ffwd submodule change": {
+            "non-ffmerge with ffwd submodule change xxxx": {
```
novalis: What's the xxxx?

node/test/util/merge_util.js
```diff
@@ -378,6 +380,7 @@ a=B:Ca-1;Cb-1;Ba=a;Bb=b|
 x=S:C2-1 s=Sa:a;C3-1 s=Sa:b;Bmaster=2;Bfoo=3`,
                 fromCommit: "3",
                 expected: `x=E:Qmessage\n#M 2: 3: 0 3;I *s=~*S:a*S:b`,
+                fails: true,
```
novalis: Why are these two tests now expected to fail?

node/lib/util/merge_util.js
shijinglu: ack

node/test/util/merge_bare.js
shijinglu: ahh, marked xxxx for debugging purpose, let me remove it

node/test/util/merge_util.js
shijinglu: Originally errors are handled in two ways in: (1) `merge_utils` returns a object that contains `errorMessage` and `merge` command throws a `UserError` if `errorMessage` is not empty. and (2) `merge_utils` throws UserError directly. 

Now all errors from the `MergeUtil` module are thrown by `MergeUtil`. This test case was the case where `MergeUtil` returns an `errorMessage` but now throws an error. Hence it was considered as not fail before but actually it fails.

PR #706, opened 2019-05-22
Turn off merge with bare in normal merge mode

shijinglu:
Previously we turned on merge with half-open repo by default. Now we change the default option to always open subrepos.

### DIFF
--- a/node/lib/cmd/merge.js
+++ b/node/lib/cmd/merge.js
@@ -173,7 +173,7 @@ Merge of '${commitName}'
                                          null,
                                          commit,
                                          mode,
-                                         Open.SUB_OPEN_OPTION.ALLOW_BARE,
+                                         Open.SUB_OPEN_OPTION.FORCE_OPEN,
                                          args.message,
                                          editMessage);
     if (null !== result.errorMessage) {


novalis:
why?

shijinglu:
> why?

Merge has three modes, `ALLOW_BARE`, `FORCE_OPEN` and `FORCE_BARE` respectfully.  `FORCE_OPEN` behaves exactly the same as the old normal merge. `FORCE_BARE` also behaves as expected and has been tested under different scenarios. `ALLOW_BARE` however this option has a bug: https://github.com/twosigma/git-meta/issues/707

Because the `ALLOW_BARE` option is only accessible from normal merge. I want to turn it off and turn it back on when the bug is fixed.

novalis:
Can we just fix #707?

On May 22, 2019 1:57:09 PM EDT, shijing <notifications@github.com> wrote:
>> why?
>
>Merge has three modes, `ALLOW_BARE`, `FORCE_OPEN` and `FORCE_BARE`
>respectfully.  `FORCE_OPEN` behaves exactly the same as the old normal
>merge. `FORCE_BARE` also behaves as expected and has been tested under
>different scenarios. `ALLOW_BARE` however this option has a bug:
>https://github.com/twosigma/git-meta/issues/707
>
>Because the `ALLOW_BARE` option is only accessible from normal merge. I
>want to turn it off and turn it back on when the bug is fixed.
>
>
>-- 
>You are receiving this because you commented.
>https://github.com/twosigma/git-meta/pull/706#issuecomment-494904271

-- 

shijinglu:
> Can we just fix #707?

This is the first step towards #707 's fix. I will put up it's fix later (fix is straightforward, but writing unit tests takes some time).

PR #708, opened 2019-06-04
Upgrade NodeGit to v0.23.0

shijinglu:


node/package-lock.json
```diff
@@ -0,0 +1,2527 @@
+{
+  "name": "git-meta",
```
motlin: Do you really want to commit the lock file? Feels like it was intentionally not committed earlier.

node/package.json
```diff
@@ -36,7 +36,7 @@
     "deeper": "",
     "fs-promise": "",
     "group-by": "",
-    "nodegit": "^0.21.1",
+    "nodegit": "^0.23.0",
```
motlin: Why ^ ? Are the floating versions the reason you added the lockfile?

node/lib/util/rm.js
```diff
@@ -194,8 +194,13 @@ const checkCleanliness = co.wrap(function *(repo, headTree, index, pathname,
         return yield checkIndexIsHeadOrWorkdir(repo, headTree, entry,
                                                pathname, displayPath);
     }
-
-    const status = yield NodeGit.Status.file(repo, pathname);
+    let status;
+    try {
+        // Status.file throws errors after 0.22.0
```
novalis: What did it do before that? And under what condition does it throw?

node/package-lock.json
shijinglu: Yes, according to the [official doc](https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md) of `package-lock.json`, it is intended to be checked in:

> package-lock.json is automatically generated for any operations where npm modifies either the node_modules tree, or package.json. It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.
> This file is intended to be committed into source repositories, and serves various purposes:

node/package.json
shijinglu: The `^` defines which version we support. The lock file basically means the what package is currently used. Ideally all patches should be supported, but lock file makes version bumps explicit, so if error happens, it is easy to track or git-bisect to find out the issue.

motlin:
The `^` caret means a range. If you drop it, we'll use the exact version
number that you wrote. I do think it's odd that we previously used ranges
plus we didn't check in the lock file. But have you considered removing all
the carets instead of checking in the lock file? It's another way to make
version upgrades deliberate without checking in a large machine-generated
file.

node/lib/util/rm.js
shijinglu: If understand correctly, it expose all errors from lib_git: https://github.com/nodegit/nodegit/pull/1468/files

> GIT_ENOTFOUND if the file is not found in the HEAD, index, and work tree, GIT_EAMBIGUOUS if `path` matches multiple files or if it refers to a folder, and -1 on other errors

shijinglu:
> The `^` caret means a range. If you drop it, we'll use the exact version number that you wrote. I do think it's odd that we previously used ranges plus we didn't check in the lock file. But have you considered removing all the carets instead of checking in the lock file? It's another way to make version upgrades deliberate without checking in a large machine-generated file.
> […](#)

Shall we follow node community's convention? For example, most popular node modules combines the version range and lock.file, for example:

lodash: [package.json](https://github.com/lodash/lodash/blob/master/package.json), [package-lock.json](https://github.com/lodash/lodash/blob/master/package-lock.json).

node-glob: [package.json](https://github.com/isaacs/node-glob/blob/master/package.json), [package-lock.json](https://github.com/isaacs/node-glob/blob/master/package-lock.json)

colors: [package.json](https://github.com/Marak/colors.js/blob/master/package.json), [package-lock.json](https://github.com/Marak/colors.js/blob/master/package.json)

node/lib/util/rm.js
novalis: So, why do we want to ignore these errors?  Shouldn't we tell the user "hey, you tried to delete a file that doesn't exist"?

shijinglu: 1. This function checks the cleanness instead of deleting a file, as stated in the comment, `An already-deleted file is always clean`. So I think it makes sense to swallow file non-exist error.
2. Main purpose of this pr is to upgrade NodeGit, IMHO,  behavioral changes should be addressed in separated pull requests.

novalis: OK, I guess let's leave it, since this does seem to make the tests pass, and I did a fair amount of testing on this code.

novalis:
For libraries, a range is useful, since it allows broader compatibility with other libraries.  For an application, I'm less certain that it's useful.  I guess it allows upgrading one library without having to manually upgrade others.  But I don't know how often that comes up.

shijinglu:
> For libraries, a range is useful, since it allows broader compatibility with other libraries. For an application, I'm less certain that it's useful. I guess it allows upgrading one library without having to manually upgrade others. But I don't know how often that comes up.

So do you agree we keep both ranges and the lock file?

novalis:
I agree to doing both of these for now.  In the future, we should consider removing ranges.

PR #709, opened 2019-06-05
fix #707: failed to set workdir correctly if submodule first opened i…

### LINKED ISSUE #707: git merge with allow_bare option failed to reset workdir in the absorbed repo
If a submodule is opened during merge, we the absorbed repo have valid workree:

```
$ cat $(git meta root)/.git/modules/foo/bar/config
[core]
<...>
     workdir = ../../../foo/bar
```

However, this is not always true in git-meta. There are occasions workdir is set to `null` even though it should not be. aka:

```
$ cat $(git meta root)/.git/modules/foo/bar/config
[core]
<...>
     workdir = null
```

The reasons is that, if the `ALLOW_BARE` option is enabled for normal merge, git meta will first try to merge the submodule inside the absorbed repo and treat it as a bare repo. And if there are any merge conflicts, git meta will then re open the subrepo. If a subrepo was first created as the bare repo, and then opened fully, git-meta won't reset the workdir and this leads to an empty submodule under `foo/bar`.

shijinglu:
git meta fails to set workdir correctly if a submodule is first open in bare mode and then open in normal mode. 

The reason is that the repo was first created with `initExt` method, which will always try to set worktree for submodules. Currently, it sets the worktree to "null" with "bare" option enabled. While correct behavior should be that the `worktree` config should not be defined at all.

### DIFF
--- a/node/test/util/submodule_config_util.js
+++ b/node/test/util/submodule_config_util.js
@@ -639,19 +639,11 @@ foo
 
     describe("initSubmoduleAndRepo", function () {
 
-        const runTest = co.wrap(function *(repo,
-                                           subRootRepo,
-                                           url,
-                                           subName,
-                                           originUrl) {
-            if (undefined === originUrl) {
-                originUrl = "";
-            }
+        /** Setup a simple meta repo with one submodule (not opened). */
+        const setupMeta = co.wrap(function *(subRootRepo, repo, url, subName) {
             const subHead = yield subRootRepo.getHeadCommit();
-            const submodule   = yield NodeGit.Submodule.addSetup(repo,
-                                                                 url,
-                                                                 subName,
-                                                                 1);
+            const submodule =
+                yield NodeGit.Submodule.addSetup(repo,url, subName, 1);
             const subRepo = yield submodule.open();
             yield subRepo.fetchAll();
             subRepo.setHeadDetached(subHead.id());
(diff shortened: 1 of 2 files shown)


node/test/util/submodule_config_util.js
```diff
@@ -722,6 +726,32 @@ foo
...
+                                                           true);
+            assert.notExists(newSub1.workdir());
+            const newSub2 =
+                yield SubmoduleConfigUtil.initSubmoduleAndRepo(originUrl,
+                                                               repo,
+                                                               subName,
+                                                               url,
+                                                               null,
+                                                               false);
+            assert.equal(path.relative(repo.workdir(),newSub2.workdir()),
```
novalis: nit: space after comma

node/test/util/merge_util.js
```diff
@@ -385,7 +385,7 @@ x=S:C2-1 s=Sa:a;C3-1 s=Sa:b;Bmaster=2;Bfoo=3`,
 Conflicting entries for submodule ${colors.red("s")}
 `,
             },
-            "conflict in submodule": {
+            "xxx conflict in submodule": {
```
novalis: what's going on here?

shijinglu: ahh, should have removed it. I used to add "xxx" in tests, so I can grep them more easily

novalis:
Why is that test failing?

shijinglu:
> Why is that test failing?

Some old flaky test, i didn't spend time investigate why it is flaky :)

PR #718, opened 2019-07-28
Exec 'pre-commit' hook of each submodule before start committing

anglee:
Exec 'pre-commit' hook of each submodule before start committing
Also added an option --no-verify for bypassing pre-commit hooks
Caveats:
* pre-commit hooks is not supported with option [file]
* pre-commit hooks is not supported with option [--interactive]

### DIFF
--- a/node/lib/cmd/commit.js
+++ b/node/lib/cmd/commit.js
@@ -84,6 +84,13 @@ have matching commits and have no new commits.`,
         constant: true,
         help: `When amending, reuse previous messages without editing.`,
     });
+    parser.addArgument(["--no-verify"], {
+        required: false,
+        action: "storeConst",
+        defaultValue: false,
+        constant: true,
+        help: `This option bypasses the pre-commit hooks.`,
+    });
     parser.addArgument(["-i", "--interactive"], {
         required: false,
         action: "storeConst",
@@ -108,10 +115,23 @@ the URL of submodules).`,
 const doCommit = co.wrap(function *(args) {
     const Commit  = require("../util/commit");
     const GitUtil = require("../util/git_util");
-    const Hook    = require("../util/hook");
+    const Hook = require("../util/hook");
 
     const repo = yield GitUtil.getCurrentRepo();
     const cwd = process.cwd();
+
+    if (!args.no_verify) {
+        try {
+            yield Commit.execSubmodulePrecommitHooks(repo,
+                                                     cwd,
+                                                     args.all,
+                                                     args.file,
+                                                     args.interactive);
+        } catch (e) {
+            console.error("Failed to commit. \n" + e.message);
+            process.exit(1);
+        }
+    }
 
     yield Commit.doCommitCommand(repo,
                                  cwd,
@@ -124,10 +144,10 @@ const doCommit = co.wrap(function *(args) {
 });
 
 const doAmend = co.wrap(function *(args) {
-    const Commit          = require("../util/commit");
-    const GitUtil         = require("../util/git_util");
-    const Hook            = require("../util/hook");
-    const UserError       = require("../util/user_error");
+    const Commit        = require("../util/commit");
+    const GitUtil       = require("../util/git_util");
+    const Hook          = require("../util/hook");
+    const UserError     = require("../util/user_error");
 
     const usingPaths = 0 !== args.file.length;
 
@@ -136,9 +156,23 @@ const doAmend = co.wrap(function *(args) {
     }
 
     const repo = yield GitUtil.getCurrentRepo();
+    const cwd = process.cwd();
+
+    if (!args.no_verify) {
+        try {
+            yield Commit.execSubmodulePrecommitHooks(repo,
+                                                     cwd,
+                                                     args.all,
+                                                     args.file,
+                                                     args.interactive);
+        } catch (e) {
+            console.error("Failed to commit. \n" + e.message);
+            process.exit(1);
+        }
+    }
 
     yield Commit.doAmendCommand(repo,
-                                process.cwd(),
+                                cwd,
                                 args.message,
                                 args.all,
                                 args.interactive,
--- a/node/lib/util/commit.js
+++ b/node/lib/util/commit.js
@@ -358,7 +358,7 @@ exports.formatEditorPrompt  = function (status, cwd) {
  * `index`.  We need to do this whenever generating a meta-repo commit because
  * otherwise, we could commit a staged commit in a submodule that would have
  * been reverted in its open repo.
- * 
+ *
  * @param {NodeGit.Repository} repo
  * @param {NodeGit.Index}      index
  * @param {Object}             submodules name -> RepoStatus.Submodule
@@ -1467,7 +1467,7 @@ exports.calculatePathCommitStatus = function (current, requested) {
  * impossible to ignore or target those commits.  We can't use them with
  * configuration changes due to the complexity of manipulating the
  * `.gitmodules` file.
- * 
+ *
  * TODO:
  *   (a) Consider allowing previously-staged commits to be included with a
  *       flag.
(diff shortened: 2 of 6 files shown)


node/lib/cmd/commit.js
```diff
@@ -105,12 +112,49 @@ the URL of submodules).`,
...
     const GitUtil = require("../util/git_util");
     const Hook    = require("../util/hook");
 
     const repo = yield GitUtil.getCurrentRepo();
+
+    if (!args.no_verify) {
+        const usingPaths = 0 !== args.file.length;
+        const usingInteractive = args.interactive;
+        if (usingPaths) {
+            console.warn(
```
novalis: 1. I would really like to handle this case, if at all possible
2. I don't want to warn someone that precommit hooks are skipped if they don't have any hooks.

node/lib/util/hook.js
```diff
@@ -67,6 +67,8 @@ exports.execHook = co.wrap(function*(repo, name, args=[]) {
         } else if (e.stdout) {
             console.log(e.stdout);
             return e.stdout;
+        } else if (e.code === 1) {
```
novalis: The code might be 2, or 143, and we should still abort the commit, right?  Also, this function previously returned a string and now it sometimes returns a string and sometimes and int, which seems really confusing.

novalis:
Oh, and some tests please?

node/lib/util/hook.js
anglee: I agree it is confusing. I made it return a number only because the JSDoc specified so. I'd be happy to change this to however you think it makes sense. Please let me know exactly what you want me to do. Thanks.

node/lib/cmd/commit.js
anglee: I just pushed a new commit that should address both of your points.
* It should handle use [files]
* It should only warn users if there is any hooks that need to be run (submodule has 'pre-commit' hook && submodule has changes to commit) but skipped.

anglee:
> Oh, and some tests please?

I added some tests.

node/lib/util/commit.js
```diff
@@ -2222,3 +2222,99 @@ it empty. You can remove the commit entirely with "git meta reset HEAD^".`);
                                        message,
                                        subMessages);
 });
+
+const submoduleHasChangesToCommit = function(status, submoduleName) {
```
novalis: This doesn't work in the case of --amend with no changes.  Actually, I think git meta just doesn't amend in this case either (which is wrong -- it needs to change the committer and commit date), so maybe it's OK that this is also broken. Probably better not to run the hooks if we're not going to run the commit.

So I think this does not need changes.  I understand that you put the hooks here because you want to run all of the hooks before running any of the commits, so there's not much you can do.

node/lib/util/commit.js
```diff
@@ -2222,3 +2222,99 @@ it empty. You can remove the commit entirely with "git meta reset HEAD^".`);
...
+
+const getSubmodulesNameForPreCommitCheck = co.wrap(function *(repo,
+                                                              repoStatus) {
+    const SubmoduleUtil = require("../util/submodule_util");
+
+    const allSubmoduleNames = yield SubmoduleUtil.getSubmoduleNames(repo);
+    return yield filterInParallel(
+        allSubmoduleNames,
+        co.wrap(function*(submoduleName) {
+            if (!submoduleHasChangesToCommit(repoStatus, submoduleName)) {
```
novalis: nit: you could simplify this code by removing the ! and moving ```return yield submoduleHasPreCommitHook(repo, submoduleName); ``` inside the if

node/lib/util/commit.js
```diff
@@ -2222,3 +2222,99 @@ it empty. You can remove the commit entirely with "git meta reset HEAD^".`);
...
+
+    const submoduleNames = yield getSubmodulesNameForPreCommitCheck(
+        repo,
+        repoStatus
+    );
+
+    if (submoduleNames.length > 0) {
+        if (interactive) {
+            console.warn(
+                "Warning. pre-commit hooks skipped when using option [-i]"
```
novalis: Could you add a comment here?  Something like "This because pre-commit hooks happen before we offer the commit message for editing, and in interactive mode, we don't know which hooks actually need to be run until the editing is done."

node/lib/util/commit.js
```diff
@@ -2222,3 +2222,99 @@ it empty. You can remove the commit entirely with "git meta reset HEAD^".`);
...
+
+const filterInParallel = co.wrap(function *(collection, asyncPredicate) {
+    const DoWorkQueue = require("../util/do_work_queue");
+
+    const booleans = yield DoWorkQueue.doInParallel(collection,
+        asyncPredicate);
+    return collection.filter((item, index) => booleans[index]);
+});
+
+const getSubmodulesNameForPreCommitCheck = co.wrap(function *(repo,
```
novalis: 1. Rename to "getSubmoduleNamesForPrecommitCheck" (note that the "s" moved)

2. You could slightly simplify the code by having this return the submodule repos instead of the names, since you're going to need them later to execute the hooks.

node/lib/util/hook.js
novalis: Figure out why it's returning a string -- does something depend on that?  If so, I guess we have to keep it a string, and fix the javadoc.  Then replace  ```if (e.code === 1)  { return 1;}``` with ```return "hook failed with no output"``` or something.  

I would be somewhat surprised if the current code worked at all, because stdio is set to 'inherit', so we should not we be able to get stdout?  And anyway we probably couldn't get it from the error object... (I can't tell for sure because the child-process-promise docs are anemic and the child-process docs area lso pretty lacking.

node/lib/cmd/commit.js
```diff
@@ -112,51 +112,27 @@ the URL of submodules).`,
...
-            console.warn(
-              "Warning. pre-commit hooks skipped when using option [file]"
-            );
-        } else if (usingInteractive) {
-            console.warn(
-              "Warning. pre-commit hooks skipped when using option [-i]"
-            );
-        } else {
-            yield execPreCommitHooks(repo);
+        try{
```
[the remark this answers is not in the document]
anglee: fixed

node/lib/util/commit.js
anglee: fixed.

node/lib/util/commit.js
anglee: 1. renamed.
2. I think returning strings make interface (return type) more flexible for client code to consume and also it is easier to stub during unit tests. Please let me know if you feel strongly that we should change this to return repos.

node/lib/util/commit.js
anglee: Added a comment.

node/lib/util/hook.js
anglee: I did a quick search in the code base and did not see the return of Hook.execHook being used. Looking at the code of `child-process-promise`:
https://github.com/patrick-steele-idem/child-process-promise/blob/master/lib/index.js
The result.status is never set and so Hook.execHook would return undefined when spawn succeeded. I suspect it meant to return `result.code`. And I agree with you that without using the [capture](https://github.com/patrick-steele-idem/child-process-promise#capture) option, the current code should not work at all.

anglee: I think fixing the bigger problem of this function is out of the scope of this PR. I pushed a change to return `"hook failed with no output"` as you suggested.

node/lib/util/commit.js
```diff
@@ -2322,7 +2322,7 @@ exports.execSubmodulePrecommitHooks = co.wrap(function *(repo,
                     submoduleName
                 );
                 const status = yield Hook.execHook(subRepo, "pre-commit");
-                if (status === 1) {
+                if (status === "hook failed with no output") {
```
novalis: We still want to stop if it fails with output (imagining that execHook were written correctly).

node/lib/util/hook.js
novalis: You are still checking that the code === 1.  Don't do that -- just unconditionally return that message.

anglee: done

node/lib/util/commit.js
anglee: ok. what would you recommend to do there? I am thinking for this function to return an object of shape: 
```
interface Result {
    code: number;
    status: string;
}
```
what do you think?

anglee: I just pushed a change to address this. The execHook function now returns a boolean indicating 'isOk'. By 'ok', we mean no error, that is, either the hook doesn't exist or hook ran successfully without errors.

PR #720, opened 2019-07-31
retry certain push failures

novalis:


node/lib/util/git_util.js
```diff
@@ -360,7 +360,13 @@ exports.push = co.wrap(function *(repo, remote, source, target, force, quiet) {
     const execString = `\
 git -C '${repo.path()}' push ${forceStr} ${remote} ${source}:${target}`;
     try {
-        const result = yield ChildProcess.exec(execString, {env : environ});
+        let result = yield ChildProcess.exec(execString, {env : environ});
+        if (result.error &&
```
shijinglu: do we need to set a max number of retries here? 
Retry without sleeping or limit make me nervous

node/lib/util/git_util.js
```diff
@@ -360,7 +360,13 @@ exports.push = co.wrap(function *(repo, remote, source, target, force, quiet) {
...
+            result.stderr.indexOf("reference already exists") !== -1) {
+            // GitLab has a race condition that somehow causes this to
+            // happen spuriously -- let's retry.
+            result = yield ChildProcess.exec(execString, {env : environ});
```
shijinglu: is on retry good enough?

novalis: I think so -- this should not happen at all, but even assuming it does, it shouldn't happen again since at the time the second push starts, the first push's ref update will already have succeeded (and failed, but the success wins)

PR #722, opened 2019-09-04
allow the synthetic ref check to exclude by path

novalis:
I manually tested just to make sure that path looked right.  This is hard to do an automated test for because it breaks some of the assumptions of git meta.

### DIFF
--- a/node/lib/util/synthetic_branch_util.js
+++ b/node/lib/util/synthetic_branch_util.js
@@ -54,14 +54,23 @@ function identity(v) {
     return v;
 }
 
-function SyntheticBranchConfig(whitelistPattern) {
-    if (whitelistPattern.length > 0) {
-        const whitelistRE = new RegExp(whitelistPattern);
-        this.whitelistTest = function(url) {
-            return whitelistRE.test(url);
+function SyntheticBranchConfig(urlWhitelistPattern, pathWhitelistPattern) {
+    if (urlWhitelistPattern.length > 0) {
+        const urlWhitelistRE = new RegExp(urlWhitelistPattern);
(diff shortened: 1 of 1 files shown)


node/lib/util/synthetic_branch_util.js
```diff
@@ -184,10 +212,15 @@ function* checkSubmodules(repo, commit) {
...
 
     const config = yield repo.config();
-    const whitelistPattern = (
+    const urlWhitelistPattern = (
         yield ConfigUtil.getConfigString(
             config, "gitmeta.skipsyntheticrefpattern")) || "";
-    const cfg = new SyntheticBranchConfig(whitelistPattern);
+    const pathWhitelistPattern = (
+        yield ConfigUtil.getConfigString(
+            config, "gitmeta.skipsyntheticrefpathpattern")) || "";
```
shijinglu: isn't this a little dangerous? An empty string will match all strings in RegEx, aka `RegEx('').test(anything)` always return true.

This behavior is different from what defined in node/lib/util/synthetic_branch_util.js:73 where if regex test always return false. So probably we can do 

```
ield ConfigUtil.getConfigString(config, "gitmeta.skipsyntheticrefpathpattern")) || ConfigUtil.DEFAULT_PATTERN;
```
here and in ConfigUtil, we can check if patten === DEFAULTPATTERN

shijinglu:
would you give some details about the manual tests? Like how `gitmeta.skipsyntheticrefpathpattern` can be configured etc?

novalis:
The manual test I did was just to print path to check that it was a non-prefixed path from the root of the repo.  Since the other regexp check worked, I assumed this one would.

PR #727, opened 2019-09-07
refuse to open unmerged submodules, like git does

novalis:


node/lib/cmd/open.js
```diff
@@ -118,6 +118,11 @@ exports.executeableSubcommand = co.wrap(function *(args) {
             return;                                                   // RETURN
         }
 
+        if (shas[index] === null) {
```
shijinglu: can we rename index here? Two index variables in the same function is really confusing.

novalis: Doing that in a separate commit.

PR #728, opened 2019-09-07
Better handling of add/add conflicts

novalis:
Well, I didn't get everything I wanted done, but at least I got this.


node/lib/util/cherry_pick_util.js
```diff
@@ -523,11 +526,49 @@ exports.computeChanges = co.wrap(function *(repo, index, targetCommit) {
...
-Conflicting entries for submodule ${colors.red(name)}
+Conflicting entries for submodule ${colors.red(name)}.
 `;
+        if (conflict.our !== null && conflict.their !== null) {
+            // add-add
+            const root = repo.workdir();
+            const our = conflict.our.id;
+            const their = conflict.their.id;
+            errorMessage += `
+To choose your version of the ${name}, use the following magic:
```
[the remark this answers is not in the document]
novalis: The third way is more complicated: we could see if the two sets share a merge-base and if so rebase inside the submodule.  That's a TODO -- it's one reason why I did the half-open.   But I would guess that this same merge-base thing only happens in some small number of cases; mostly add/adds are e.g. different people importing the same ext codebase.

People might do something stupid, and if so it'll show up in code review.  So hopefully someone will notice.

node/lib/util/test_util.js
```diff
@@ -281,3 +281,7 @@ exports.makeBareCopy = co.wrap(function *(repo, path) {
     yield NodeGit.Remote.delete(bare, "origin");
     return bare;
 });
+
+exports.quotemeta = function(str) {
```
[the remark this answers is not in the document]
novalis: Because we're now doing regex matching.

node/lib/util/cherry_pick_util.js
```diff
@@ -523,11 +526,49 @@ exports.computeChanges = co.wrap(function *(repo, index, targetCommit) {
...
+            const bare = Open.SUB_OPEN_OPTION.FORCE_BARE;
+            const subRepo = yield opener.getSubrepo(name, bare);
+
+            for (const stage of [conflict.ancestor, conflict.our,
+                                 conflict.their]) {
+                if (stage !== null) {
+                    yield fetcher.fetchSha(subRepo, name, stage.id);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
novalis: The question is: what sha?  Here, we loop over the stages, which is what's needed in this case.  But is that really generally useful?

novalis:
(just removed a few console.logs)

node/lib/util/cherry_pick_util.js
```diff
@@ -523,11 +526,47 @@ exports.computeChanges = co.wrap(function *(repo, index, targetCommit) {
...
+    const opener = new Open.Opener(repo, null);
+    const fetcher = yield opener.fetcher();
     for (let name of names) {
-        yield ConflictUtil.addConflict(index, name, conflicts[name]);
+        let conflict = conflicts[name];
+        let configured = false;
+        try {
+            // We just want to see if it's configured
+            const url = yield fetcher.getSubmoduleUrl(name);
+            configured = url !== null;
```
bpeabody: There's nothing in the contract of `getSubmoduleUrl` that says it might return a `null`; what's the check for?

novalis: Well, it definitely happens in the tests...

bpeabody: I can't see how it could return a null after a (quick) look at the implementation, but if it can the contract should be updated to explain when; as it is the contract is to throw a `UserError` on failure.

novalis: if this.d_urls[name] actually contains null, that would happen. I think this may only happen in tests.  I could change the code to instead detect that case and throw UserError (then catch it here instead of checking for null) -- what do you think?

bpeabody: I'd want to understand how `this.d_urls` comes to have null values first, and what it means.

novalis: Oh, wait a second -- I think I just checked the wrong thing.  In fact it's not null. I could just set configured to true here.

PR #729, opened 2019-09-13
Write notes for all parents and simplify the synthetic branch check

novalis:
Simplify the synthetic branch check no longer recurse in parallel.

Recusing in parallel can create exponentially-many threads; it seems
to be extremely ineffecient.

Writing for all parents makes future lookups faster.

### DIFF
--- a/node/lib/util/synthetic_branch_util.js
+++ b/node/lib/util/synthetic_branch_util.js
@@ -271,6 +271,9 @@ function* checkSubmodules(repo, commit) {
  * Returns true if the check passes on all branches; false if any
  * fail.
  *
+ * On success, create a note reflecting the work done to save time
+ * on future updates.
+
  * @async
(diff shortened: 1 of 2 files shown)


node/lib/util/synthetic_branch_util.js
```diff
@@ -318,12 +318,20 @@ function* parentLoop(repo, notesRepo, commit, oldSha, handled) {
...
 
     const parents = yield commit.getParents(commit.parentcount());
-    const parentChecks = yield parents.map(function *(parent) {
-        return yield *parentLoop(repo, notesRepo, parent, oldSha, handled);
-    });
-    const result = parentChecks.every(identity);
-    handled[commit.id()] = result;
-    return result;
+    let success = true;
+    for (const parent of parents) {
```
shijinglu: could you update the doc of `parantLoop` to match the update behavior

PR #731, opened 2019-09-24
set exec's maxBuffer

novalis:
It's kind of annoying that we have to do this, but I've hit this a few
times for reasons I don't really understand. So let's just make the
buffer huge and hope we don't fill it.

### DIFF
--- a/node/lib/util/git_util.js
+++ b/node/lib/util/git_util.js
@@ -48,6 +48,8 @@ const Hook                = require("./hook");
 const GitUtilFast         = require("./git_util_fast");
 const UserError           = require("./user_error");
 
+const EXEC_BUFFER = 1024*1024*100;
+
 /**
  * Create a branch having the specified `branchName` in the specified `repo`
  * pointing to the current head.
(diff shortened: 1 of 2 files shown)


node/lib/util/git_util.js
```diff
@@ -360,12 +362,18 @@ exports.push = co.wrap(function *(repo, remote, source, target, force, quiet) {
     const execString = `\
 git -C '${repo.path()}' push ${forceStr} ${remote} ${source}:${target}`;
     try {
-        let result = yield ChildProcess.exec(execString, {env : environ});
+        let result = yield ChildProcess.exec(execString, {
+            env : environ,
+            maxBuffer: EXEC_BUFFER
+        });
```
shijinglu: there are plenty of similar objects, you may want to simplify them with object spread operator. For example:

```
baseOpt = {maxBuffer: 1024*1024*100}
...
let result = yield ChildProcess.exec(execString, {...baseOpt, env: environ})
```

novalis: That's not that much shorter, so I think I'll skip it for now.

PR #734, opened 2019-10-14
Make push start with submodule's workdir if one exists

anglee:
This change would allow pre-push hooks using Husky to run correctly. Before this change, submodule pushes were ran from the bare repos (in meta-repo's `.git/modules`) and husky could not be found with the path (specified in .git/modules/../hooks). Making push command run from the workdir (ie. not from `.git`) makes husky run correctly.

### DIFF
--- a/node/lib/util/push.js
+++ b/node/lib/util/push.js
@@ -236,7 +236,7 @@ exports.push = co.wrap(function *(repo, remoteName, source, target, force) {
         const sha = pushMap[subName];
         const syntheticName =
                           SyntheticBranchUtil.getSyntheticBranchForCommit(sha);
-        const subRepo = yield SubmoduleUtil.getBareRepo(repo, subName);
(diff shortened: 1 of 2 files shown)


node/lib/util/push.js
```diff
@@ -236,7 +236,7 @@ exports.push = co.wrap(function *(repo, remoteName, source, target, force) {
         const sha = pushMap[subName];
         const syntheticName =
                           SyntheticBranchUtil.getSyntheticBranchForCommit(sha);
-        const subRepo = yield SubmoduleUtil.getBareRepo(repo, subName);
+        const subRepo = yield SubmoduleUtil.getRepo(repo, subName);
```
anglee: My understanding is that with getRepo, it get's the submodule repo with workdir() set (while using getBareRepo gets the repo in the meto repo's `.git/modules` which workdir() is undefined). Thus, the expectation here is that with getRepo, `subRepo` will have workdir() set and below git_util.push can then use the workdir().

PR #738, opened 2019-11-15
Do not fail to push when a changed submodule is closed

novalis:
All we need is the bare repo to push.

This bit a user inside Two Sigma, and I tested this fix in his repo and it works.


novalis:
[I manually cherry-picked this to master because Craig isn't official an approver but is officially my boss and thus has the right to approve this]

novalis:
Apparently the manual cherry-pick didn't work

PR #739, opened 2020-01-06
fix merge-bare no-ff param behavior not be inverted

chaimmintz:
Previously the --no-ff param behavior was inverted where "--no-ff" allowed fast-forwards and the absense of "--no-ff" forced a commit. This fixes it to work as expected.


novalis:
This one wants a test.

novalis:
After talking with Chaim, since this is a bug in the driver, it's a hassle to test and unlikely to regress.

PR #748, opened 2020-01-14
create synthetic ref for fetches in submodules so that subsequent fet…

chaimmintz:
…ches can use it for negotiation and only download new objects instead of the entire history


novalis:
Looks like you'll have to fix up the test harness to ignore these new synthetic refs.

PR #749, opened 2020-01-15
Add error message to remind user to use git meta push

shijinglu:
pre receive hook can reminder user of using `git meta push` when it deteches submodule sysnthetic branches are not updated along with the meta commit user is trying to push.

### DIFF
--- a/node/lib/util/synthetic_branch_util.js
+++ b/node/lib/util/synthetic_branch_util.js
@@ -395,7 +395,12 @@ function* metaUpdateIsBad(repo, notesRepo, updates) {
     });
 
     const resolved = yield checkFailures;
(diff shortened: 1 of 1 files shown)


novalis:
I would like to make this message more prominent -- for instance, by adding some newlines before and after it.  Also, can you discuss how you tested it?

shijinglu:
> I would like to make this message more prominent -- for instance, by adding some newlines before and after it. Also, can you discuss how you tested it?

Thanks for the reminder, just updated and added new lines.

The bad meta update case is already tested in the test/synthetic-branch.js, so the test plan is that all existing tests should pass.  Adding an extra test for console.log seems not that critical.

PR #752, opened 2020-02-28
Allow merge-bare to best effort merge submodule urls

shijinglu:
merge-bare failes when merge `.gitmodules` conflicts. The change in this PR deals with the problem in two steps:
1. inside `computeChangesBetweenTwoCommits`, compute urls of .gitmodules by doing a three way merge and return the urls.
2.  in bare merge mode, instead update urls in `CherryPickUtil.changeSubmodules`, we directly write the post-merge ursl to index.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "6.9.1"
+  - "8.11.1"
 
 env:
   - CC=gcc-5
--- a/node/lib/util/merge_util.js
+++ b/node/lib/util/merge_util.js
@@ -52,6 +52,7 @@ const SubmoduleChange     = require("./submodule_change");
 const SubmoduleFetcher    = require("./submodule_fetcher");
 const SubmoduleRebaseUtil = require("./submodule_rebase_util");
 const SubmoduleUtil       = require("./submodule_util");
+const SubmoduleConfigUtil = require("./submodule_config_util");
 const UserError           = require("./user_error");
 
 const CommitAndRef    = SequencerState.CommitAndRef;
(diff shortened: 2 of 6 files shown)


node/lib/util/merge_util.js
```diff
@@ -370,6 +371,7 @@ const mergeStepPrepare = co.wrap(function *(context) {
         return MergeStepResult.error(errorMessage);                  // RETURN
     }
 
+    // TODO: remove this
```
novalis: Do we want to remove this now, or?

novalis: I would rather have users hit errors than have PRS hit errors.

shijinglu: actually that is a good idea. Updated: 

Originally `changeSubmodules` (non-bare) does two things, (1) delete/open submdoules according to differences bettween urls read from merged .gitmodules vs urls read from HEAD.  (2) writes the urls into .gitmodules on HEAD.

Now I move the second step outside of `changeSubmodules`, and wirte the "smart merged" urls into .gitmodules just like merge-bare would do.

shijinglu: ok changeSubmodules is used by a number of places, so needs more work

shijinglu: now, passing the urls to changeSubmodules

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "6.9.1"
+  - "8.11.1"
```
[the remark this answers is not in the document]
shijinglu: it was because with node v6.9.1 and the orginal `package.json`. Travis CI always fail (https://travis-ci.org/twosigma/git-meta/builds/656330522?utm_medium=notification&utm_source=github_status). I suspect it is because some public library was force updated with the same version. To make travis ci pass and to embrace newer NodeJs, I updated the node version

PR #753, opened 2020-03-03
Express fetch error during merging more explicitly

shijinglu:


node/lib/util/merge_util.js
```diff
@@ -230,9 +230,17 @@ exports.mergeSubmodule = co.wrap(function *(metaIndex,
...
-    if (null !== change.ourSha) {
-        yield fetcher.fetchSha(subRepo, subName, change.ourSha);
+    try {
+        yield fetcher.fetchSha(subRepo, subName, theirSha);
+        if (null !== change.ourSha) {
+            yield fetcher.fetchSha(subRepo, subName, change.ourSha);
+        }
+    } catch (e) {
+        console.error(
+            `Unable to fetch changes in submodule '${subName}', ` +
```
novalis: Did we say that we were going to put this error to stdout?  I'm worried about parsing stderr with stuff from git fetch mixed in.

shijinglu: Use error because it is an error :)

PR #754, opened 2020-03-04
Revert node_js version bump

shijinglu:
Previously node install fails instantly without bumping the package-lock.json. And I think it is better just bump the node version, so we can be benifited from newer es6 syntax. However, some legacy platforms seems does not support it :(

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "8.11.1"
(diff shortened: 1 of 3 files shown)


novalis:
also need to revert use of ... syntax

shijinglu:
> also need to revert use of ... syntax

that syntax comes from a dependency :(, let me ping the package to its older version

PR #757, opened 2020-04-01
If our sha is a descendant of their sha, we need to make sure that our

novalis:
sha gets added to the index.  This is handled elsewhere in non-bare merges,
but in bare merges, we need to manually handle it.


novalis:
g

PR #758, opened 2020-04-01
add option to merge{,-bare} to read commit message from a file

novalis:


node/lib/cmd/merge_bare.js
```diff
@@ -97,10 +103,13 @@ exports.executeableSubcommand = co.wrap(function *(args) {
     const MergeUtil      = require("../util/merge_util");
     const MergeCommon    = require("../util/merge_common");
     const GitUtil        = require("../util/git_util");
-    const Hook           = require("../util/hook");
     const Open           = require("../util/open");
     const UserError      = require("../util/user_error");
 
+    if (Boolean(args.message) + Boolean(args.message_file) !== 1) {
+        throw new UserError("Use exactly one of -m and -F.");
+    }
```
shijinglu, line 111: a little strange this is different from the check in `merge.js`

novalis: for merge.js, neither is required (if you omit them, your editor will pop up).  Here, before this change, -m was required by the argparse bit.

I could have written them in a more similar way, I guess, but I was going for easy-to-read rather than similar.

PR #763, opened 2020-06-03
submodule commit hook failures should be silent

novalis:


novalis:
That's what regular git doesb

PR #766, opened 2020-07-07
Provides momre information when merge fails:

shijinglu:
1. do not crash even if a submodule is corruptted and its HEAD points to a nonexistent commit
2. if merge failed in a submodule, print what submodule it was


novalis:
Can we make this two separate commits?

PR #768, opened 2020-07-14
Fix up-to-date case in git merge-bare.

shijinglu:
In a nomral merge (merge "their" commit into "ours"), if "ours" is a descendant of "theirs" then it is considered as up-to-date case. If "theirs" is a descedant of "ours", then it is a fast-forward merge.

In merge-bare mode, "theirs" and "ours" are symetrical, and git-meta should not exit early as if it is an up-to-date case

### DIFF
--- a/node/lib/util/merge_util.js
+++ b/node/lib/util/merge_util.js
@@ -402,13 +402,6 @@ const mergeStepPrepare = co.wrap(function *(context) {
         return MergeStepResult.justMeta(infoMessage, theirCommit);    // RETURN
     }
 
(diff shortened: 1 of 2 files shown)


shijinglu:
> Shouldn't the case of merging an ancestor in succeed with the output "Already up to date."? That's what regular git merge does.

Good catch, that log was dropped. Adding it back based on

https://github.com/twosigma/git-meta/blob/9358cf2e5d83a7f8540023795e3b0c4da541e298/node/lib/util/merge_util.js#L317

shijinglu:
discussed offline, it is probably better to keep merge-bare and merge behave the same

PR #770, opened 2020-07-22
Write temp index and give pre-commit hooks access to read and modify it

novalis:
This is a giant hassle, as you can imagine.

### DIFF
--- a/node/lib/cmd/commit.js
+++ b/node/lib/cmd/commit.js
@@ -120,25 +120,13 @@ const doCommit = co.wrap(function *(args) {
     const repo = yield GitUtil.getCurrentRepo();
     const cwd = process.cwd();
 
-    if (!args.no_verify) {
-        try {
-            yield Commit.execSubmodulePrecommitHooks(repo,
-                                                     cwd,
-                                                     args.all,
-                                                     args.file,
-                                                     args.interactive);
-        } catch (e) {
-            console.error("Failed to commit. \n" + e.message);
-            process.exit(1);
-        }
-    }
-
     yield Commit.doCommitCommand(repo,
                                  cwd,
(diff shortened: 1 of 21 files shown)


novalis:
BTW, fs.copyFileSync works on my machine so I have no clue why travis sucks.  Will fix it.

node/lib/cmd/commit.js
```diff
@@ -120,25 +120,13 @@ const doCommit = co.wrap(function *(args) {
     const repo = yield GitUtil.getCurrentRepo();
     const cwd = process.cwd();
 
-    if (!args.no_verify) {
-        try {
-            yield Commit.execSubmodulePrecommitHooks(repo,
```
shijinglu, line 125: `Commit.execSubmodulePrecommitHooks` is removed, isn't it? If so, could you also remove it at line 163?

node/lib/util/git_util.js
```diff
@@ -997,3 +997,29 @@ exports.getReference = co.wrap(function *(repo, name) {
...
     return null;
 });
+
+/**
+ * Read an index file, and overwrite an in-memory index with its contents.
+ *
+ * @param {NodeGit.Index} index
+ * @param {String|undefined} path if set, whence to read the index
+ */
+exports.overwriteIndexFromFile = co.wrap(function*(index, path) {
```
shijinglu, line 1007: why not just open a new index and replace the old index instance?

novalis: Because the old index instance is what repo.index() returns, and there are places that call that to get the index.

node/lib/cmd/commit.js
novalis: thanks

PR #771, opened 2020-07-27
Precommit

novalis:


node/lib/util/commit.js
```diff
@@ -2198,14 +2217,15 @@ exports.doCommitCommand = co.wrap(function *(repo,
         return yield exports.commitPaths(repo,
                                          repoStatus,
                                          message,
-                                         subMessages);
+                                         noVerify);
```
novalis, line 2219: This never worked.  I hate JS.

PR #774, opened 2020-08-03
run post-open hooks after opening subs in reset and cherry-pick

novalis:


shijinglu:
Are there any other places that needs to call post-open hook? For example git meta commit also tries to open submodules, do we need to change that as well?

novalis:
> Are there any other places that needs to call post-open hook? For example git meta commit also tries to open submodules, do we need to change that as well?

So, I am not sure what the right thing to do is.  

One idea is to copy-paste this code into yet more places.  That's not ideal, because what if we add a new one and forget to paste again?

Another thing we could do is to change the opener code, so that every time a submodule is opened, the hook is executed on that individual submodule.  But if you then git meta open . on a repo with many (10k, say) submodules, that's a lot of exec calls.  Do we care?  It's arguably better than what we are doing now, which runs the risk of hitting command-line argument length limits (which the current system does risk, although Two Sigma is probably a few years out from hitting it).

We could also change the opener code so that we cover all of the cases like commit, reset, etc, but have a special case for git meta open which runs the hook only once.

What do you think?

shijinglu:
IMHO, changing the opener is the right thing to do, and I don't see a good reason why we need to keep git meta open as a special case:
1. As you said, many exec calls is probably better than single call with long argument. 
2. It makes a lot sense to keep open related logic to the opener.
3. It will be a maintenance burden to remember to run post-hook wherever repo is opened. Especially when adding new features, for example git meta diff.

PR #777, opened 2020-08-19
Allow cherry-picking multiple commits.

novalis:


node/lib/cmd/cherry_pick.js
```diff
@@ -120,22 +119,25 @@ exports.executeableSubcommand = co.wrap(function *(args) {
...
-    const annotated = yield GitUtil.resolveCommitish(repo, commitish);
-    if (null === annotated) {
-        throw new UserError(`\
+    // TOOD: check if we are mid-rebase already
+
+    const commits = args.commit;
+
+    const resolvedCommits = [];
+    for (let commitish of commits) {
+        const annotated = yield GitUtil.resolveCommitish(repo, commitish);
```
shijinglu, line 128: to my understanding the multiple commits implies that the arg here is a "commitish" that should be resolved with git rev-list, not necessarily a list of commits, right? 

For example: `maint master..next` in vanilla git mean 
>all commits that are ancestors of maint or next, but not master or any of its ancestors.

Does `GitUtil.resolveCommitish` resolves commitish like git rev-list does?

node/lib/util/cherry_pick_util.js
```diff
@@ -779,9 +817,9 @@ exports.rewriteCommit = co.wrap(function *(repo, commit) {
  * @return {Object}      returm.submoduleCommits
  * @return {String|null} return.errorMessage
  */
-exports.cherryPick = co.wrap(function *(metaRepo, commit) {
+exports.cherryPick = co.wrap(function *(metaRepo, commits) {
```
shijinglu, line 820: could you also update doc with `@param commits` ?

node/lib/util/cherry_pick_util.js
```diff
@@ -805,22 +843,23 @@ running cherry-pick.`);
     // cherry-pick file.
 
     const head = yield metaRepo.getHeadCommit();
-    const seq = new SequencerState({
+    const commitIdStrs = [];
+    let lastCommit;
+    for (const commit of commits) {
```
shijinglu, line 848: nit: you can use map if commits is an array, like
```
commits.map(x => x.id().tostrS() )
```

node/lib/util/cherry_pick_util.js
```diff
@@ -755,21 +757,57 @@ exports.rewriteCommit = co.wrap(function *(repo, commit) {
...
     return result;
 });
 
+const pickRemainingCommits = co.wrap(function*(metaRepo, seq) {
+    const commits = seq.commits;
+    let result;
+    for (let i = seq.currentCommit; i < commits.length; ++i) {
+        const id = commits[i];
+        const commit = yield metaRepo.getCommit(id);
+        console.log(`Cherry-picking commit ${colors.green(id)}.`);
```
shijinglu, line 775: don't you need to check if there are conflicts before applying the rewriting the commit?

node/lib/cmd/cherry_pick.js
novalis: I think you're right -- that is the way that regular git cherry-pick does it.

That said, this is what the Two Sigma user requested, and it is an improvement that we could merge.  I would do the rev-list thing as a follow-up.

node/lib/util/cherry_pick_util.js
novalis: see above where it checks status.isConflicted

PR #778, opened 2020-08-19
Synthetic check

novalis:
check that synthetic refs exist as well as the commits they point to.


novalis:
Don't approve this -- I think we should probably drop it.

PR #781, opened 2020-08-25
Let git-meta notice changes to .gitmodules

shijinglu:
With the change:
- git meta status will show changes to .gitmodules
- git meta add . will stage .gitmodules
- git meta commit -a -m will include .gitmodules changes

### DIFF
--- a/node/lib/util/diff_util.js
+++ b/node/lib/util/diff_util.js
@@ -35,7 +35,6 @@ const co      = require("co");
 const NodeGit = require("nodegit");
 
 const RepoStatus          = require("./repo_status");
-const SubmoduleConfigUtil = require("./submodule_config_util");
 
 const DELTA = NodeGit.Diff.DELTA;
 
@@ -81,11 +80,8 @@ function readDiff(diff) {
                      delta.newFile();
         const path = file.path();
 
-        // Skip the .gitmodules file and all submodule changes; they're handled
-        // separately.
-
-        if (SubmoduleConfigUtil.modulesFileName !== path &&
-            NodeGit.TreeEntry.FILEMODE.COMMIT !== file.mode()) {
+        // Skip the all submodule changes; they're handled separately.
+        if (NodeGit.TreeEntry.FILEMODE.COMMIT !== file.mode()) {
             result[path] = fileStatus;
         }
     }
--- a/node/lib/util/status_util.js
+++ b/node/lib/util/status_util.js
@@ -356,6 +356,116 @@ exports.readConflicts = function (index, paths) {
 };
 
 /**
+ * Return the status of submodules.
+ *
+ * @async
+ * @param {NodeGit.Repository} repo
+ * @param {Object}             [options] see `getRepoStatus` for option fields
+ * @param {NodeGit.Commit}     headCommit HEAD commit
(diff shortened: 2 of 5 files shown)


node/lib/util/status_util.js
```diff
@@ -465,8 +465,19 @@ exports.getRepoStatus = co.wrap(function *(repo, options) {
                                                     options.paths,
                                                     options.ignoreIndex,
                                                     options.showAllUntracked);
-        args.staged = status.staged;
-        args.workdir = status.workdir;
+        // if showMetaChanges is off, keep .gitmodules changes only
```
novalis, line 468: Hm, is that really the right thing to do?  isn't a .gitmodules change a meta change?

novalis: Also, this function is getting really long -- would it be possible to refactor somehow?

node/test/util/commit.js
```diff
@@ -1183,6 +1183,9 @@ a=B:Chi#a-1;Ba=a|x=U:C3-2 s=Sa:a;Bmaster=3;Os W README.md=888`,
                     status: new RepoStatus({
                         currentBranchName: "master",
                         headCommit: "2",
+                        staged: {
```
novalis, line 1186: I don't expect this change, since I don't see .gitmodules mentioned in the state.  Why is this happening?  (Indeed, I don't expect .gitmodules to ever be deleted by the actions of git-meta).

node/test/util/diff_util.js
```diff
@@ -353,6 +353,9 @@ x=S:C2 a/b=c,a/c=d,t=u;H=2;I a/b,a/q=r,f=x;W a/b=q,a/c=f,a/y=g,f`,
                 input: `
 a=B:Ca-1;Bmaster=a|
 x=S:C2-1 s=Sa:1;Bmaster=2`,
+                staged: {
+                    ".gitmodules": FILESTATUS.ADDED,
```
novalis, line 357: ... or added.

node/test/util/commit.js
shijinglu: For this test, it deletes the only submodule "s", and hence deleting .gitmodules is reasonable. But most importantly, it is not this PR  that deletes the .gitmodules, the PR only make changes to .gitmodules visible to git-meta.

node/test/util/diff_util.js
shijinglu: same as the above, this PR does not add .gitmodules, but it makes changes to .gitmodules visible to git-meta status

node/lib/util/diff_util.js
```diff
@@ -83,9 +82,7 @@ function readDiff(diff) {
 
         // Skip the .gitmodules file and all submodule changes; they're handled
```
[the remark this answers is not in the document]
shijinglu: good catch

node/lib/util/status_util.js
shijinglu: let me see what I can do

node/test/util/commit.js
novalis: I guess I'm surprised that someone thought of that case, but maybe it's libgit2 that did so, which would be less surprising.

PR #786, opened 2020-09-10
optimize isOpen for the common case, to not read all submodule dirs

novalis:


node/lib/util/open.js
```diff
@@ -254,12 +256,20 @@ Opener.prototype.clearAbsorbedCache = function (subName) {
...
  */
-Opener.prototype.isOpen = co.wrap(function *(subName) {
-    if (!this.d_initialized) {
-        yield this._initialize();
+Opener.prototype.isOpen = function (subName) {
+    if (this.d_initialized) {
+        return this.d_openSubs.has(subName) || (subName in this.d_cachedSubs);
+    } else {
+        const modulesDir = path.join(this.d_repo.path(), "modules");
+        const submodulePath = path.join(modulesDir, subName);
```
shijinglu, line 264: does  .git always reference HEAD of half opened modules?

shijinglu (approved):
Looks straightforward, but would be nice to have a short explanation why x/y/z is considered open if
`.git/module/x/y/x/HEAD` exist and `x/y/z/.git` are present

node/lib/util/open.js
novalis: This method is asking whether a submodule is fully-open not half-open. Half-open submodules are prevented by the check a few lines down.

PR #787, opened 2020-09-14
Reject amend commits mid merge/rebase/cherry-pick

novalis:


node/lib/cmd/commit.js
```diff
@@ -148,6 +149,13 @@ const doAmend = co.wrap(function *(args) {
     const repo = yield GitUtil.getCurrentRepo();
     const cwd = process.cwd();
 
+    const seq = yield SequencerStateUtil.readSequencerState(repo.path());
+    if (seq) {
+        const ty = seq.type.toLowerCase();
+        const msg = "You are in the middle of a " + ty + " -- cannot amend.";
+        throw new UserError(msg);
+    }
+
```
shijinglu, line 158: can we move the logic to `Commit.doAmendCommand`, and I think we can even make a util function to check if git meta is in the middle of a merge/rebase/cherry-pick.

novalis: I considered that, but it doesn't actually look cleaner to me.  readSequencerState is just one call, and we need the seq.type for the error message.  Plus in the third (not-yet-implemented) case, I think we need seq.target.

shijinglu: Have you considered just use the NodeGit API instead of `readSequencerState` to determine the repos state? 
https://www.nodegit.org/api/repository/#state can tells you if an operation (merge, cherry-pick, etc) is in progress, and the return is state type, like `MERGE`, `CHERRY_PICK`, `BISECT` etc.

novalis: meta has its own sequencer which is independent of the regular git sequencer.

shijinglu (approved):

PR #788, opened 2020-09-16
make commit --all mid-merge continue the merge

novalis:


node/lib/util/commit.js
```diff
@@ -497,7 +498,8 @@ exports.commit = co.wrap(function *(metaRepo,
                                     metaStatus,
                                     message,
                                     subMessages,
-                                    noVerify) {
+                                    noVerify,
+                                    mergeParent) {
```
shijinglu, line 502: could you also update the doc with the new parameter, and and a short description on what happens if `mergeParent` is present?

node/lib/util/commit.js
```diff
@@ -550,6 +558,13 @@ exports.commit = co.wrap(function *(metaRepo,
             if (headCommit !== null) {
                 parents.push(headCommit);
             }
+
+            if (mergeTree !== null) {
+                const mergeSubParent = yield mergeTree.entryByPath(name);
+                if (mergeSubParent.id() !== headCommit.id()) {
+                    parents.push(mergeSubParent.id());
```
shijinglu, line 565: in line 559, we just pushed a commit to `parents`, now we are pushing an object id to `parents`. Is it designed to have different types in the same array?

node/lib/util/commit.js
```diff
@@ -597,14 +612,22 @@ exports.commit = co.wrap(function *(metaRepo,
 
     const tree = yield index.writeTree();
     const headCommit = yield metaRepo.getHeadCommit();
+    const parents = [headCommit];
+    if (mergeParent) {
+        parents.push(mergeParent);
```
shijinglu, line 617: what happens if `mergeParent.id() === headCommit.id()`? Will it be two parents of the same commit?

node/test/util/commit.js
```diff
@@ -3385,3 +3385,74 @@ x=U:Cmsg\n#x-2 s=Sa:s;Bmaster=x;Os Cmsg\n#s-1 README.md=foo,addedbyhook=bar`;
...
+            yield MergeUtil.merge(repo,
+                                  null,
+                                  m4,
+                                  MergeCommon.MODE.NORMAL,
+                                  Open.SUB_OPEN_OPTION.FORCE_OPEN,
+                                  [],
+                                  "msg",
+                                  function() {});
+        } catch (e) {
+            // we expect the merge to fail, but that's OK
```
shijinglu, line 3419: why is merge fail expected? Any chance we can create a test case that merge can finish successfully?

node/lib/util/commit.js
novalis: It passes the tests, and it didn't when I accidentally pushed mergeSubParent (without the .id()).  So. apparently, yes.

node/lib/util/commit.js
novalis: I think git meta merge prevents you from starting a merge like that, but I can add an assert.

node/test/util/commit.js
novalis: This is testing the case that you are mid-merge (that is, the merge has failed and you are now doing conflict resolution).

shijinglu (approved):

PR #789, opened 2020-09-22
Use git instead of node-git to determine what commit's submodules need pushing

novalis:


node/lib/util/push.js
```diff
@@ -66,38 +67,46 @@ const EMPTY_TREE = "4b825dc642cb6eb9a060e54bf8d69288fbee4904";
  *
  * @async
  * @param {NodeGit.Repository} repo
- * @param {String}             remoteName
  * @param {NodeGit.Commit}     commit
  */
-exports.getClosePushedCommit = co.wrap(function*(repo, remoteName, commit) {
```
shijinglu, line 72: could you also update the comment? The comment still talks about param `remoteName` can be used

node/lib/util/push.js
```diff
@@ -66,38 +67,46 @@ const EMPTY_TREE = "4b825dc642cb6eb9a060e54bf8d69288fbee4904";
...
         }
-        throw err;
     }
+    if (excludeRefs.length === 0) {
+        return null;
+    }
+
+    const execString = `\
+git -C '${repo.workdir()}' rev-list --reverse --topo-order ${commit} \
+${excludeRefs.join(" ")}`;
```
shijinglu, line 93: do you want to limit the number of revs to return?  like `-n 1`?

novalis: Yes, but unfortunately, I can't, because -n happens before --reverse.

shijinglu (approved):

PR #791, opened 2020-09-25
Fix generate-repo.js.

alexst2:
Looks like the RepoAST API changed and now requires `RepoAST.File` instances to be passed.


novalis (approved):
Thanks!

PR #792, opened 2020-09-25
avoid exceptions from calling git-meta checkout on unopened submodules

tsuikevi-twosigma:
Running git meta checkout throws an error if it's passed a path to an unopened submodule e.g.

```
~/P/g/foo [master]× » gm checkout -- .
Error: could not find repository from '/Users/kevintsui/Projects/git-meta/foo/au/gp/unm'
```

This change avoids this error by not applying the checkout operation to unopened submodules.

### DIFF
--- a/node/lib/util/checkout.js
+++ b/node/lib/util/checkout.js
@@ -601,7 +601,11 @@ function noSuchFileMessage(subName, path) {
 exports.checkoutFiles = co.wrap(function*(repo, options) {
     assert.instanceOf(repo, NodeGit.Repository);
     const resolvedPaths = options.resolvedPaths;
-    const subNames = Object.keys(resolvedPaths);
+
+    // Exception is thrown if we try to get repo info from unopened submodules
+    const openSubmodules = yield SubmoduleUtil.listOpenSubmodules(repo);
+    const subNames = Object.keys(resolvedPaths).filter(
+        subName => openSubmodules.includes(subName));
 
     let subCommits;
     let stage = 0;
--- a/node/lib/util/checkout.js
+++ b/node/lib/util/checkout.js
@@ -602,10 +602,13 @@ exports.checkoutFiles = co.wrap(function*(repo, options) {
     assert.instanceOf(repo, NodeGit.Repository);
     const resolvedPaths = options.resolvedPaths;
 
-    // Exception is thrown if we try to get repo info from unopened submodules
+    // Exception is thrown if we try to get repo info from unopened submodules.
+    // TODO: handle other use cases besides when a commit is not specified.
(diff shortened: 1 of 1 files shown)


node/lib/util/checkout.js
```diff
@@ -601,7 +601,11 @@ function noSuchFileMessage(subName, path) {
 exports.checkoutFiles = co.wrap(function*(repo, options) {
     assert.instanceOf(repo, NodeGit.Repository);
     const resolvedPaths = options.resolvedPaths;
-    const subNames = Object.keys(resolvedPaths);
+
+    // Exception is thrown if we try to get repo info from unopened submodules
```
novalis, line 605: Shouldn't we open the submodules instead?  Or, if the path is the entire submodule, I think we should be able to check out without opening the submodule.

tsuikevi-twosigma: With the example I gave above `gm checkout -- .`, opening submodules doesn't sound like the right thing to do because it would result in every submodule being opened. I'm not sure how we can do a checkout on files without opening the submodule, can you explain what that would look like?
`checkout` with files as an argument is a working tree operation and opening a submodule to modify the working tree seems like the wrong thing to do; opening submodules as a side-effect of running `checkout` doesn't sound desirable.

novalis: We can checkout without opening the submodule by just changing where the submodule pointer in the meta repo's index points.  You can actually see this using regular git.  Just run `git checkout $rev -- foo` on some closed submodule.  I should note that I am not 100% sure that `git meta open` respects this, but if it doesn't, that's a bug.  `git meta commit` definitely should

tsuikevi-twosigma: Ah ok I see your point about modifying the submodule pointer. Covering the use cases you described is going to take me more time which I won't have for a while.
What do you think about special casing this behaviour so that we only ignore unopened submodules if `$rev` is empty or `HEAD`? In these cases, ignore the unopened submodules is ok because they are already pointing to the correct commit.
With this approach, we can at least improve the UX for certain use cases now and I could address the other use cases in a future hack day.

novalis: That seems fine for now.  I guess it's technically wrong in the case where someone has already used regular git checkout to change the submodule pointers to something else, but that's pretty rare.

tsuikevi-twosigma: Ok great, I'll work on updating the pr.

tsuikevi-twosigma:
Ended up only covering the case where `$rev` is empty. Checking for `HEAD` was doable but not as clean as I would have liked for this partial fix so decided not to implement it.

novalis (approved):

Issue #711, opened 2019-06-20
Improve error message when -i and --amend --no-edit are passed together


petroseskinder:
In vanilla git, one cannot pass `-i` and `--amend --no-edit`. The error response with vanilla git is 

```
fatal: Only one of --include/--only/--all/--interactive/--patch can be used.
```
With `git-meta` the error message is much less clear
```
git meta commit -ai --amend --no-edit
TypeError: editMessage is not a function
    at Object.<anonymous> (/.../lib/util/commit.js:2154:32)
    at Generator.next (<anonymous>)
    at onFulfilled (/.../node_modules/co/index.js:65:19)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

As expected, when the `-i` or `-no-edit` is not passed `git-meta` works as expected. We should improve git meta's error message to reflect git's error messages.

petroseskinder:
I think we just need to add more requirements in the `commit` argument parsing
https://github.com/twosigma/git-meta/blob/abce63fc037827653a0c2a11d70ea51ecc6731ec/node/lib/cmd/commit.js#L71

novalis:
Fixed in #794

PR #794, opened 2020-09-25
Show error when committing with -i and -a flags

kaaniboy:
Issue #711 shows that running `git meta commit -ia --amend --no-edit` causes an unhandled error. In vanilla git, the `-i` (interactive) and `-a` (all) options are not allowed together in the first place:

```
▶ git commit -ia
fatal: Only one of --include/--only/--all/--interactive/--patch can be used.
```

This PR adds a  check to the `commit` command's argument parsing to prevent `-i` and `-a` from being used together:

```
▶ gm commit -ia
The use of '-i' and '-a' does not make sense.
```

### DIFF
--- a/node/lib/cmd/commit.js
+++ b/node/lib/cmd/commit.js
@@ -193,8 +193,8 @@ exports.executeableSubcommand = function (args) {
         process.exit(1);
     }
     if (args.all && args.interactive) {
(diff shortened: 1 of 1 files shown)


novalis (approved):

novalis (approved):

novalis:
(this was merged but github didn't notice)

PR #797, opened 2020-09-29
Commit msg

novalis:
This one is on top of #796

### DIFF
--- a/node/lib/util/commit.js
+++ b/node/lib/util/commit.js
@@ -201,7 +201,7 @@ exports.stageChange = co.wrap(function *(index, path, change) {
 
 const mutex = new Mutex();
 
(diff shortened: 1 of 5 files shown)


node/lib/util/commit.js
```diff
@@ -198,16 +199,20 @@ exports.stageChange = co.wrap(function *(index, path, change) {
     }
 });
 
-const runHooks = co.wrap(function *(repo, index) {
+const mutex = new Mutex();
```
shijinglu, line 202: i think this is mixed with the other PR, could you rebase?

shijinglu (commented):
Could you elaborate what is the purpose of this change in the commit message (or the commit title)? 
And maybe squash the commits as well?

shijinglu (approved):

PR #803, opened 2020-12-04
Add --force-with-lease to push.

### LINKED ISSUE #732: FR: `git meta push` should support --force-with-lease
This is much safer than the currently-supported `--force`.

alexst2:
This only implements --force-with-lease as a boolean flag, i.e. it does not support specifying ref names yet.

Fixes #732.

### DIFF
--- a/node/lib/cmd/push.js
+++ b/node/lib/cmd/push.js
@@ -32,6 +32,8 @@
 
 const co = require("co");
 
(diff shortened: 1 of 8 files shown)


novalis (approved):

novalis:
Thanks, this looks great!

PR #807, opened 2020-12-16
Reduce ref negotiation during stitching

novalis:
We turn off negotiation entirely the first time we fetch a submodule.  Thereafter, we only negotiate for refs from that submodule.

This contains two other minor changes: one removes an unused parameter and the other removes some useless code.


novalis:
I tested this inside Two Sigma (as well as running the unit tests)

shijinglu (approved):

PR #813, opened 2021-01-22
Make commit faster with many open submodules

novalis:
I manually tested this inside Two Sigma on the repo that our user
complained about, and it seems to work.

### DIFF
--- a/node/lib/util/commit.js
+++ b/node/lib/util/commit.js
@@ -387,12 +387,44 @@ exports.formatEditorPrompt  = function (status, cwd) {
  * @param {Object}             submodules name -> RepoStatus.Submodule
  */
 const stageOpenSubmodules = co.wrap(function *(repo, index, submodules) {
-    yield Object.keys(submodules).map(co.wrap(function *(name) {
+    const entries = yield Object.keys(submodules).map(co.wrap(function *(name) {
         const sub = submodules[name];
         if (null !== sub.workdir) {
-            yield index.addByPath(name);
+            /*
+             We probably shouldn't run addByPath in parallel because
+             one of the following two things is probably true:
+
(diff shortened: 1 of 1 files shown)


node/lib/util/commit.js
```diff
@@ -387,12 +387,44 @@ exports.formatEditorPrompt  = function (status, cwd) {
...
+            const newEntry = new NodeGit.IndexEntry();
+            newEntry.flags = 0;
+            newEntry.flagsExtended = 0;
+            newEntry.mode = NodeGit.TreeEntry.FILEMODE.COMMIT;
+            newEntry.id = head;
+            newEntry.path = name;
+            return newEntry;
         }
     }));
+    for (const entry of entries.filter(x => x)) {
```
shijinglu, line 425: nit: it is more common to use `x => !!x` to filter out undefined or empty in javascript.

node/lib/util/commit.js
```diff
@@ -387,12 +387,44 @@ exports.formatEditorPrompt  = function (status, cwd) {
...
+             lock means slow submodule HEAD lookups are effectively
+             serialized despite this being a parallel map.
+
+             So instead, we'll do the submodule HEAD lookups in
+             parallel and then add the entries serially, which should
+             be relatively fast.
+            */
+
+            const subRepo = yield SubmoduleUtil.getRepo(repo, name);
+            const head = yield subRepo.getHeadCommit();
```
shijinglu, line 415: do we need to worry about HEAD missing case?

shijinglu (approved):

node/lib/util/commit.js
novalis: I don't think that's a real case -- the submodule has a workdir, so it almost certainly has a HEAD which points to some commit.

PR #815, opened 2021-01-22
Test for cherry-pick

novalis:
I missed this as part of 72057ba42a6fa3184a1c6f4b30373a41e9d28cf1

### DIFF
--- a/node/test/util/cherry_pick.js
+++ b/node/test/util/cherry_pick.js
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2021, Two Sigma Open Source
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


node/test/util/cherry_pick.js
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright (c) 2020, Two Sigma Open Source
```
shijinglu, line 2: 2021?

shijinglu (approved):

node/test/util/cherry_pick.js
novalis: I wrote it in 2020 but forgot to check it in then.

novalis: oops, actually it should be 2021 -- publication date.  will fix.

PR #818, opened 2021-02-02
workAroundLibgit2MergeBug to skip creating synthetic ancestor if ance…

shijinglu:
…stor has the same sha as theirs.

In the [submoudle_change.js](https://github.com/twosigma/git-meta/blob/master/node/lib/util/submodule_change.js#L56), submodule change with same sha is forbidden, as I qoute:

```
    /**
     * Creat a new `Changed` object having the specified `oldSha` and `newSha`
     * values.  The behavior is undefined if `oldSha === newSha`
    ....
     */
    constructor(oldSha, newSha, ourSha) {
        assert.notEqual(oldSha, newSha);
```

However,  workAroundLibgit2MergeBug routine  can still create this "empty" change object with same newSha as the oldSha.

Let's add a logic to exlude that edge cases

### DIFF
--- a/node/lib/util/cherry_pick_util.js
+++ b/node/lib/util/cherry_pick_util.js
@@ -282,10 +282,12 @@ const workAroundLibgit2MergeBug = co.wrap(function *(data, repo, name,
         for (const base of mergeBases) {
             const shas = yield SubmoduleUtil.getSubmoduleShasForCommit(
                 repo, [ours.path], base);
-            if (shas[name] !== undefined) {
-
+            const sha = shas[name];
+            // Avoid creating a synthetic ancestor with the same sha as
+            // theirs. See more in `SubmoduleChange`
+            if (sha !== undefined && sha !== theirs.id.tostrS()) {
                 ancestor = new NodeGit.IndexEntry();
-                ancestor.id = NodeGit.Oid.fromString(shas[name]);
+                ancestor.id = NodeGit.Oid.fromString(sha);
                 ancestor.mode = NodeGit.TreeEntry.FILEMODE.COMMIT;
                 ancestor.path = ours.path;
                 ancestor.flags = ours.flags;


node/lib/util/submodule_util.js
```diff
@@ -610,7 +610,9 @@ exports.resolveSubmoduleNames = function (workdir,
         const relPath = GitUtil.resolveRelativePath(workdir,
                                                     cwd,
                                                     filename);
-        const result = exports.getSubmodulesInPath(relPath, submoduleNames, true);
```
novalis, line 613: what's going on with these changes?  They look like just formatting?

node/yarn.lock
```diff
@@ -184,6 +184,14 @@ buffer-fill@^1.0.0:
   resolved "https://registry.yarnpkg.com/buffer-fill/-/buffer-fill-1.0.0.tgz#f8f78b76789888ef39f205cd637f68e702122b2c"
   integrity sha1-+PeLdniYiO858gXNY39o5wISKyw=
 
+call-bind@^1.0.0, call-bind@^1.0.2:
```
novalis, line 187: I don't think we need to change yarn.lock?

novalis (changes requested):
I would like to see a test here, if possible.

node/lib/util/submodule_util.js
shijinglu: yes, somehow they total line length exceeds 80 but not complained by travis ci...

shijinglu:
> I would like to see a test here, if possible.

Tested it with two sigma's proprietary monorepo, could not find a way to replicate that test case here. But the change is pretty safe. The worst case senario is that workAroundLibgit2MergeBug is skipped and user is left with submodule add-add conflict. In that case, use can just resolve the submodule conflict manually.

novalis (approved):

PR #820, opened 2021-02-05
format

novalis:


node/lib/util/submodule_util.js
```diff
@@ -610,7 +610,8 @@ exports.resolveSubmoduleNames = function (workdir,
         const relPath = GitUtil.resolveRelativePath(workdir,
                                                     cwd,
                                                     filename);
-        const result = exports.getSubmodulesInPath(relPath, submoduleNames, true);
+        const result = exports.getSubmodulesInPath(relPath, submoduleNames,
```
shijinglu, line 613: Most of the code formats this like:

```
const result = exports.getSubmodulesInPath(relPath,
                                           submoduleNames,
                                           true)
```
Do we want to keep that?

shijinglu (approved):

PR #821, opened 2021-03-02
Make stash support commits inside submodules

novalis:


node/lib/util/stash_util.js
```diff
@@ -148,30 +151,159 @@ exports.save = co.wrap(function *(repo, status, includeUntracked, message) {
...
+                // meta index, and new commits in the submodule (which
+                // may or may not be related to those staged in the
+                // index).
+
+                // In theory, our data structures support writing this case,
+                // but since we don't yet support reading it, we probably
+                // shouldn't let the user get into a state that they can't
+                // easily get out of.
+
+                throw new UserError(`BUG: ${name} is in a state that is too \
```
shijinglu, line 213: probably should not call this a bug

node/lib/util/stash_util.js
```diff
@@ -148,30 +151,159 @@ exports.save = co.wrap(function *(repo, status, includeUntracked, message) {
...
+                                          sig,
+                                          sig,
+                                          null,
+                                          "stash",
+                                          indexTree,
+                                          4,
+                                          [subCommit,
+                                           indexCommit,
+                                           indexCommit,
+                                           indexCommit]);
```
shijinglu, line 197: why does this commit have 4 parents and three of them are the same?

node/lib/util/stash_util.js
```diff
@@ -310,30 +443,84 @@ exports.apply = co.wrap(function *(repo, id, reinstateIndex) {
...
+            } catch (e) {
+                // We expect these errors to be caught, but if
+                // something goes wrong, wWe are leaving the user in a
+                // pretty yucky state.  the alternative is to try to
+                // back out the whole stash apply, which seems worse.
+
+                rebaseError(`We are leaving the rebase half-finished, so \
+that you can fix the conflicts and continue by running 'git rebase \
+--continue' inside of ${name} (note: not 'git meta rebase').`);
+                console.error(`The underlying error, which is likely useless \
```
shijinglu, line 517: I think user usually is not interested in knowing about nodegit, would you make the message more user friendly, maybe something like "unkown error but unlikely harmful"?

node/test/util/stash_util.js
```diff
@@ -338,6 +338,11 @@ x=E:Fmeta-stash=s;
                     }
                 }
 
+                for (const c of Object.keys(commitMap)) {
```
shijinglu, line 341: why is this needed?

novalis: In some cases, the first or second parent might be an existing commit, but the test framework only wants commitMap to contain commits it has never seen, so we have to remove the ones that already existed. I will add a comment to this effect.

node/lib/util/stash_util.js
novalis: I'll edit the docs.

node/lib/util/stash_util.js
novalis: (changed)

node/lib/util/stash_util.js
novalis: (changed)

shijinglu (approved):

PR #823, opened 2021-03-16
Support an alternate submodule repository for stitching

novalis:


node/lib/util/stitch_util.js
```diff
@@ -961,13 +962,31 @@ exports.stitch = co.wrap(function *(repoPath,
...
+         * meta repository commits).  It must be an alternate
+         * of this repository (the repository that stitching is
+         * being done in), so that we can access the objects
+         * that we fetch to it.  This lets us have a second alternate
+         * repository for just the meta repository commits.  And that
+         * saves a few seconds on meta repository fetches.
+         */
+        const subsRepoPath = yield ConfigUtil.getConfigString(
+            config,
+            "gitmeta.stitchSubmodulesRepository");
```
shijinglu, line 979: The function is exported (public), so could you also mention this config string in the function doc?

shijinglu (approved):

node/lib/util/stitch_util.js
novalis: I don't think that makes sense?  It's an internal detail of the stitching process, so callers of the outer method probably don't need to be aware of it.  I added a doc instead.

shijinglu (approved):

PR #824, opened 2021-03-16
Fail git meta open if you request a sub that does not exist

novalis:


node/lib/cmd/open.js
```diff
@@ -156,7 +156,8 @@ exports.executeableSubcommand = co.wrap(function *(args) {
     let subsToOpen = yield SubmoduleUtil.resolveSubmoduleNames(workdir,
                                                                cwd,
                                                                subs,
-                                                               paths);
+                                                               paths,
+                                                               true);
```
shijinglu, line 160: to me this is a bit hacky. IMO, it is better to return full result, like a map of sub names to actually opened codebases, and caller can then decide whether to and throw exception if need to. But I can imagine that will introduce plenty of changes.

shijinglu (approved):

PR #825, opened 2021-04-08
Fail git meta close if you request a sub that does not exist, instead…

novalis:
… of closing the current sub(!)


shijinglu (commented):
did you test the change, could you explain what the bug was?

novalis:
Added a test, and updated the commit message

shijinglu (approved):

PR #826, opened 2021-04-15
Use better notes ref update method

novalis:
First, add retries.

But second, there was a TOCTOU race in updating.  In order to fix
this, I first tried to use existing NodeGit methods, but they call
into a function of libgit2 which didn't support asserting that a ref
was new.  I filed a PR against libgit2 to fix this, but of course it
will take some time to get merged.  So in the meantime, we shell out
to git.

### DIFF
--- a/node/lib/util/bulk_notes_util.js
+++ b/node/lib/util/bulk_notes_util.js
@@ -31,6 +31,7 @@
 "use strict";
 
 const assert         = require("chai").assert;
+const ChildProcess   = require("child-process-promise");
 const co             = require("co");
 const NodeGit        = require("nodegit");
 const path           = require("path");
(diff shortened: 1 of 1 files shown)


node/lib/util/bulk_notes_util.js
```diff
@@ -108,9 +105,64 @@ exports.writeNotes = co.wrap(function *(repo, refName, contents) {
                                                newTree,
                                                parents.length,
                                                parents);
-    yield NodeGit.Reference.create(repo, refName, commit, 1, "updated");
+
+    let old;
+    if (null === ref) {
+        old = "0000000000000000000000000000000000000000";
+    } else {
+            old = ref.target().tostrS();
```
shijinglu, line 113: nit: too many spaces

node/lib/util/bulk_notes_util.js
```diff
@@ -53,37 +54,33 @@ exports.shardSha = function (sha) {
...
+    // note containing `contents`.
 
     let currentTree = null;
     const parents = [];
     const ref = yield GitUtil.getReference(repo, refName);
     if (null !== ref) {
         const currentCommit = yield repo.getCommit(ref.target());
         parents.push(currentCommit);
-        currentTree = yield currentCommit.getTree();
+            currentTree = yield currentCommit.getTree();
```
shijinglu, line 83: nit: remove extra leading spaces

shijinglu (approved):

PR #827, opened 2021-04-27
stitch: Handle missing converted commit

novalis:
Handle the case where a notes is present but the converted commit to is not
present.  This can happen if you are stitching and pushing notes in parallel:
it's possible that you push a note for a commit that doesn't end up getting
pushed.  That would cause later conversions which fetch the notes to fail,
unless we handle that case.

### DIFF
--- a/node/lib/util/stitch_util.js
+++ b/node/lib/util/stitch_util.js
@@ -778,8 +778,15 @@ exports.readConvertedCommits = co.wrap(function *(repo) {
 
     const result =
                  yield BulkNotesUtil.readNotes(repo, exports.convertedNoteRef);
-    for (let key in result) {
-        result[key] = exports.readConvertedContent(result[key]);
+    for (const [key, oldSha] of Object.entries(result)) {
+        const sha = exports.readConvertedContent(oldSha);
+        try {
+            yield repo.getCommit(sha);
+            result[key] = sha;
+        } catch (e) {
+            // We have the note but not the commit, delete from cache
+            delete result[key];
+        }
     }
     return result;
 });
@@ -804,6 +811,14 @@ exports.makeGetConvertedCommit = function (repo, cache) {
             return cache[sha];
         }
         const result = yield exports.readConvertedCommit(repo, sha);
+
+        try {
+            yield repo.getCommit(result);
+        } catch (e) {
+            // We have the note but not the commit; treat as missing
+            return undefined;
+        }
+
         cache[sha] = result;
         return result;
     });


node/lib/util/stitch_util.js
```diff
@@ -778,8 +778,15 @@ exports.readConvertedCommits = co.wrap(function *(repo) {
...
-        result[key] = exports.readConvertedContent(result[key]);
+    for (let key of Object.keys(result)) {
+        const sha = exports.readConvertedContent(result[key]);
+        try {
+            yield repo.getCommit(sha);
+        } catch (e) {
+            // We have the note but not the commit, delete from cache
+            delete result[key];
+        }
+        result[key] = sha;
```
shijinglu, line 789: `result[key] = sha;` will always be executed, right? Do you want to move this inside try block?

node/lib/util/stitch_util.js
```diff
@@ -778,8 +778,15 @@ exports.readConvertedCommits = co.wrap(function *(repo) {
 
     const result =
                  yield BulkNotesUtil.readNotes(repo, exports.convertedNoteRef);
-    for (let key in result) {
-        result[key] = exports.readConvertedContent(result[key]);
+    for (let key of Object.keys(result)) {
```
shijinglu, line 781: nil, `for (const [key, sha] of Object.entries(result))` is probably easier to read

node/lib/util/stitch_util.js
```diff
@@ -778,8 +778,15 @@ exports.readConvertedCommits = co.wrap(function *(repo) {
...
+        const sha = exports.readConvertedContent(result[key]);
+        try {
+            yield repo.getCommit(sha);
+        } catch (e) {
+            // We have the note but not the commit, delete from cache
+            delete result[key];
```
shijinglu, line 787: the outside loops over keys and inside the loop, it changes the key set, this pattern reminds me of the ConcurrentModificationException in java. Is it too expensive to return a new object instead of modifying the existing object?

node/lib/util/stitch_util.js
```diff
@@ -804,6 +811,14 @@ exports.makeGetConvertedCommit = function (repo, cache) {
             return cache[sha];
         }
         const result = yield exports.readConvertedCommit(repo, sha);
+
+        try {
+            yield repo.getCommit(result);
+        } catch (e) {
+            // We have the note but not the commit; treat as missing
+            return undefined;
```
shijinglu, line 819: Will it be a problem if those commits are not converted? I think they maybe pushed later, right?

node/lib/util/stitch_util.js
novalis: oops

node/lib/util/stitch_util.js
novalis: No, it loops over Object.keys(result). That's computed first (not lazily).

node/lib/util/stitch_util.js
novalis: We'll just reconvert them.

shijinglu (approved):

PR #834, opened 2021-07-12
Consider the index when comparing the tree to worktree during a commit.

novalis:
It turns out that libgit2 has a function for that, which is good, because
the alternative involved much painful hacking of git_tree_to_workdir.

### DIFF
--- a/node/test/util/commit.js
+++ b/node/test/util/commit.js
@@ -3449,12 +3449,20 @@ describe("commit mid-merge", function() {
         const index = yield sRepo.index();
         const sEntry = index.getByIndex(1);
         const newEntry = new NodeGit.IndexEntry();
-        newEntry.flags = sEntry.flags;
+        // mask off the stage bits here, so that the new entry we write is
+        // in stage zero.
+        newEntry.flags = sEntry.flags & ~0x3000;
         newEntry.flagsExtended = sEntry.flagsExtended;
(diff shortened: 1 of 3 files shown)


node/test/util/commit.js
```diff
@@ -3449,12 +3449,20 @@ describe("commit mid-merge", function() {
         const index = yield sRepo.index();
         const sEntry = index.getByIndex(1);
         const newEntry = new NodeGit.IndexEntry();
-        newEntry.flags = sEntry.flags;
+        // mask off the stage bits here, so that the new entry we write is
+        // in stage zero.
+        newEntry.flags = sEntry.flags & ~0x3000;
```
leahnp, line 3454: can you explain this one to me?

novalis: We're trying, here, to mark a conflict as resolved (never mind that there are still conflict markers -- this is just a test).  Once a conflict is resolved, the stage should be 0.   The stage bits are 0x3000.  We want everything *but* the stage bits, so we bitwise-and with the bitwise-inverse of 0x3000.  So imagine the original bits are 0x6abc.  0x6abc & 0xcfff gives 0x4abc  -- that is, everything but the stage bits are the same, and the stage bits are zero.  So now the entry's stage is zero, which I think is what `git add` does.

leahnp (approved):

PR #838, opened 2021-08-05
Stash deletion of submodules

novalis:


node/lib/util/stash_util.js
```diff
@@ -178,34 +247,42 @@ report this.  Continuing stash anyway.`);
...
-            // This is a case that regular git stash doesn't really have
-            // to handle. In a normal stash commit, the tree points
-            // to the working directory tree, but here, there is no working
-            // directory.  But if there were, we would want to have
-            // this commit checked out.
-
-            const subRepo = yield SubmoduleUtil.getRepo(repo, name);
+            if (sub.index === null || sub.index.sha === null) {
+                // deleted submodule
+                stashId = MAGIC_DELETED_SHA;
```
leahnp, line 252: does this make a special stash, that would be different than a user made stash?

novalis: It's different in the meta repo, but not in the submodule repo.  Since regular stashes in the meta repo are not really a thing that does anything, I think this is OK.

leahnp:
This is quite complex, I think I understand it and reviewed it the best I can - I am happy to approve unless you want another set of eyes on it.

leahnp (approved):

node/lib/util/repo_status.js
```diff
@@ -372,7 +372,8 @@ class Submodule {
      */
     isCommittable () {
         return !(this.isNew() &&
-                 null === this.d_index.sha &&
+                 (null === this.d_index ||
+                  null === this.d_index.sha) &&
```
shijinglu, line 376: nit: you can use optional chaining: `?.`

node/lib/util/rm.js
```diff
@@ -461,7 +461,21 @@ exports.rmPaths = co.wrap(function *(repo, paths, options) {
...
+                            // Repo still exists for some reason --
+                            // perhaps it was reported as deleted
+                            // because it's not in the index, but it
+                            // does exist on disk.  For safety, do not
+                            // delete it; this is a weird case.
+                            console.error(
+                                `Could not remove ${fullpath} -- it's not
+empty.  If you are sure it is not needed, you can remove it yourself.`
+                            );
+                        }
```
shijinglu, line 477: do you want to raise exception if `("ENOTEMPTY" !== e.code)`? Otherwise the exception will be swallowed, right?

node/lib/util/stash_util.js
```diff
@@ -107,6 +117,63 @@ exports.makeLogMessage = co.wrap(function *(repo) {
...
+
+    const result = {};
+    // RepoStatus throws away the diff new sha, and rather than hack
+    // it, since it's used all over the codebase, we'll just redo the
+    // diffs for this one file.
+
+    const workdirToTreeDiff =
+          yield NodeGit.Diff.treeToWorkdir(repo,
+                                           headTree,
+                                           {paths: [GITMODULES]});
```
shijinglu, line 150: It calculates `treeToWorkdir`, which implies it always `includeUntracked`, right? 

How do you handle the `includeUntracked` parameter in `exports.save`?

novalis: This only applies to .gitmodules, which should never be untracked?

node/lib/util/repo_status.js
novalis: Not with this version of jshint :(

PR #839, opened 2021-08-06
replace travis with github actions

novalis:
This is still flaky, but the flakiness is inherent to the test (that is, I also see it locally), and I don't really understand why it's flaky. 

Also there was some weird test that failed in github actions but not locally, so I hacked in what should be a reasonable fix.


leahnp:
can you link to the tests you ran in github actions?

novalis:
https://github.com/novalis/git-meta/actions/runs/1105222607

leahnp (approved):

PR #847, opened 2021-08-20
Stash deletion of submodules (second try)

novalis:


node/lib/util/stash_util.js
```diff
@@ -107,6 +117,62 @@ exports.makeLogMessage = co.wrap(function *(repo) {
...
+
+    const result = {};
+    // RepoStatus throws away the diff new sha, and rather than hack
+    // it, since it's used all over the codebase, we'll just redo the
+    // diffs for this one file.
+
+    const workdirToTreeDiff =
+          yield NodeGit.Diff.treeToWorkdir(repo,
+                                           headTree,
+                                           {pathspec: [GITMODULES]});
```
novalis, line 149: change since last version: s/paths/pathspec/ -- we used the wrong variable name, leading to slooooow stashes.

Issue #713, opened 2019-06-25
git meta doesn't respect color.status config


petroseskinder:
git meta status does not respect the `color.status` config. Apparently it is hard coded in the `colors/safe` npm module. 

Apparently `FORCE_COLOR=0 git meta status` works

petroseskinder:
this hit @lihalite recently

PR #851, opened 2021-12-03
Add color.status config support to address #713

clsater:


clsater:
hmm... not sure why that test is failing on the CI server -- it works when I run locally.

PR #857, opened 2022-02-28
Improve logging to list unconverted ancestors of given commit SHA

bjornbjornsson:
Previously, stitch_util.js was logging the message `listing all unconverted ancestors of {commit_sha}`, but the unconverted ancestors weren't actually being logged. This PR puts the unconverted ancestors in the log and also logs how many unconverted ancestors there are.

### DIFF
--- a/node/lib/util/stitch_util.js
+++ b/node/lib/util/stitch_util.js
@@ -873,6 +873,8 @@ exports.listCommitsToStitch = co.wrap(function *(repo,
     }
     const commitShas = exports.listCommitsInOrder(commit.id().tostrS(),
                                                   allParents);
+    console.log(`listing all ${commitShas.length} unconverted ancestors of ` +
+                `${commit.id.tostrS()}: ${commitShas}`);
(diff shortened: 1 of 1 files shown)


node/lib/util/stitch_util.js
```diff
@@ -873,6 +873,8 @@ exports.listCommitsToStitch = co.wrap(function *(repo,
     }
     const commitShas = exports.listCommitsInOrder(commit.id().tostrS(),
                                                   allParents);
+    console.log(`listing all ${commitShas.length} unconverted ancestors of ` +
+                `${commit.id.tostrS()}: ${commitShas}`);
```
shijinglu, line 877: There could be a number of commits in `commitSha`, let's truncate it if it is too long. 
You can do:
1. for each sha, take only the beginning 7 chars (full length long sha to short-sha format)
2. join array with comma
3. truncate if the total `length` exceeds 1K

bjornbjornsson: Good point, will do

PR #880, opened 2023-01-20
Run CI against Node 18

geofft:
This is the first version that officially supports OpenSSL 3.


geofft:
Internal change to switch to Node 18: https://tsdev.app.twosigma.com/pr/cr39679

